PR #5, opened 2015-09-15
update macros for magfield rescale setting

mccumbermike:
best to wait for the next new build


mccumbermike:
We have a new build so I'm adding the handles for the mag field rescaling.

PR #7, opened 2015-11-30
Introduce 4 Tracking Options using Cylinder Geos at Updated Radii and Efficiencies

mccumbermike:
Something to discuss at tomorrow's meeting. This pull request will do a few things:

(1) add 4 non-default cylinder based macros for the 4 tracking combinations. This includes hit efficiency estimates and update radial positions for the 13/19 pixel ladder reconfiguration. The outer strips have been updated with the pCDR geometry values. I also add an outer TPC cage and move the inner TPC cage to radius 30 cm. All four options execute without errors, but I haven't looked at the output for the TPC macros. Review and tweaks from technology advocates will be needed.
(2) some general clean up wherein I complete the encapsulation of global variable defines into the subsystem macros
(3) I also retire some unmaintained CEMC and HCAL macros.

By default nothing in the logic changes for the tracking.

The purpose of this merge will be to allow updated DCA and tracking performance figures for the pCDR based geometries for general use. Once the tracking software has been fully updated to accept complex geometries (help needed!) we can switch over to those non-cylinder implementations. This is to get us started.


mccumbermike:
I've also updated the beam pipe description macro for the aluminum sections at large rapidity.

mccumbermike:
Merging with 1.4 T field adjustments, 1.4T is now the default in the steering macro.

PR #10, opened 2015-12-17
fsPHENIX

johnlajoie:
added macros for fsPHENIX


macros/g4simulations/G4Setup_fsPHENIX.C
```diff
@@ -0,0 +1,286 @@
...
+  if (do_magnet) MagnetInit();
+
+  gROOT->LoadMacro("G4_HcalOut_ref.C");       // default 
+  if (do_hcalout) HCalOuterInit();
+
+  gROOT->LoadMacro("G4_Global.C");
+  
+  gROOT->LoadMacro("G4_Jets.C");
+
+  gROOT->LoadMacro("G4_FwdJets.C");
```
blackcathj: Are we missing G4_FwdJets.C in this pull request? @johnlajoie

johnlajoie: Good point. I was using a modified version of G4_Jets.C. I think having a separate G4_FwdJets.C is a good idea.  I can certainly add one.

PR #11, opened 2015-12-27
read new default mapping file for FEMC

johnlajoie:


blackcathj:
The change affect on-going forward development only. Merging it. 

Please note update to the calibration file (https://github.com/sPHENIX-Collaboration/calibrations/pull/4) would take another nightly-build to show up on RCF.

PR #12, opened 2016-01-16
Enable Calorimeter Truth Compression

mccumbermike:
I add 2 function calls to the setup macros to run the removal of compressed information and the removal of redundant nodes during the DST output.

### DIFF
--- a/macros/g4simulations/Fun4All_G4_fsPHENIX.C
+++ b/macros/g4simulations/Fun4All_G4_fsPHENIX.C
@@ -77,6 +77,8 @@ int Fun4All_G4_fsPHENIX(
   bool do_FHCAL_twr = true; 
   bool do_FHCAL_cluster = true; 
 
+  bool do_dst_compress = true;
+  
   //Option to convert DST to human command readable TTree for quick poke around the outputs
   bool do_DSTReader = true;
   //---------------
--- a/macros/g4simulations/Fun4All_G4_fsPHENIX.C
+++ b/macros/g4simulations/Fun4All_G4_fsPHENIX.C
@@ -233,6 +233,8 @@ int Fun4All_G4_fsPHENIX(
   if (do_FHCAL_twr) FHCAL_Towers();
   if (do_FHCAL_cluster) FHCAL_Clusters();
 
+  if (do_dst_compress) G4Compress();
+  
   //--------------
   // SVTX tracking
   //--------------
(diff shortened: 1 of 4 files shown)


blackcathj:
Thanks for this update. It works well for me, with this minor modification (and uncomment for the compressed DST)

```diff
diff --git a/macros/g4simulations/Fun4All_G4_sPHENIX.C b/macros/g4simulations/Fun4All_G4_sPHENIX.C
index 0a67f69..a224257 100644
--- a/macros/g4simulations/Fun4All_G4_sPHENIX.C
+++ b/macros/g4simulations/Fun4All_G4_sPHENIX.C
@@ -286,6 +286,7 @@ int Fun4All_G4_sPHENIX(
     }
 
   // Fun4AllDstOutputManager *out = new Fun4AllDstOutputManager("DSTOUT", outputFile);
+  //  if (do_dst_compress) DstCompress(out);
   // se->registerOutputManager(out);

```

mccumbermike:
Done, I somehow removed that line before committing. It is back now.

pinkenburg:
By now it is probably better to change this to selecting the nodes you 
want to save instead of having an ever growing list of nodes which are 
excluded. The list to save is probably already shorter than the 
exclusions. It would probably possible to add some wildcards for the 
nodenames you want to save, so as long as we have a naming convention 
you get away with just a few lines.

On 1/17/2016 11:26 AM, Michael P. McCumber wrote:
>
> Done, I somehow removed that line before committing. It is back now.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/sPHENIX-Collaboration/macros/pull/12#issuecomment-172347033>.
>

-- 
*************************************************************

Christopher H. Pinkenburg	;    pinkenburg@bnl.gov
				;    http://www.phenix.bnl.gov/~pinkenbu

Brookhaven National Laboratory	;    phone: (631) 344-5692
Physics Department Bldg 510 C	;    fax:   (631) 344-3253
Upton, NY 11973-5000

*************************************************************

mccumbermike:
I think I disagree. The cost of forgetting to blacklist something out of a
DST file and the compression module is the extra space is consumed in the
output file which is still valid for evaluation. The cost of forgetting to
whitelist something into a DST is an incomplete event record that may not
contain the information of a newly added non-alorimeter volume and it won't
necessarily evaluate correctly.

I constructed the showers so that they will summarize new volumes
automatically, but the g4hits themselves are preserved all the way into the
DST until they are explicitly added to the compression and stripped from
the DST. I think I prefer this failure mode to a whitelist based system
despite the extra macro lines.

We can discuss it more on Tuesday.


--
Michael P. McCumber, PhD
Los Alamos National Laboratory
505-709-8161

mccumbermike:
We agreed to merge this with the non-default setting. I'm doing that now.

PR #13, opened 2016-03-15
allow easy switch between fsPHENIX and EIC EMCal

johnlajoie:
These macro changes include an easy switch between the fsPHENIX and EIC forward EMCal implementations.  Right now this is at the level of commenting in/out different sections.

### DIFF
--- a/macros/g4simulations/G4_FEMC.C
+++ b/macros/g4simulations/G4_FEMC.C
@@ -34,11 +34,15 @@ FEMCSetup(PHG4Reco* g4Reco, const int absorberactive = 0)
 
   ostringstream mapping_femc;
 
-  /* path to central copy of calibrations repositry */
-  mapping_femc << getenv("OFFLINE_MAIN") << "/share/calibrations";
-  mapping_femc << "/ForwardEcal/mapping/towerMap_FEMC_v005.txt";
+  // EIC ECAL
+  //femc->SetEICDetector(); 
+  //mapping_femc << getenv("OFFLINE_MAIN") << "/share/calibrations/ForwardEcal/mapping/towerMap_FEMC_v005.txt";
+
+  // fsPHENIX ECAL
+  femc->SetfsPHENIXDetector(); 
+  mapping_femc << getenv("OFFLINE_MAIN") << "/share/calibrations/ForwardEcal/mapping/towerMap_FEMC_fsPHENIX_v001.txt";
+
   cout << mapping_femc.str() << endl;
-  //mapping_femc << "towerMap_FEMC_latest.txt";
 
   femc->SetTowerMappingFile( mapping_femc.str() );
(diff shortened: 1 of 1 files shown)


blackcathj:
Hi, @johnlajoie 

After coming back from Iowa, I have been trying to wrap up merging the tracking part with the calorimeter updates. I just realized a infliction that the forward field return required a diameter-19.8-cm hole in the EMCal in order to complete the field loop. Right now the piston field return cuts into the inner blocks of the re-stacked MPC as illustrated below. If the MPC mapping file would be updated with a 10x10 block hole in the middle, it should resolve this problem. 

2016-03-17 18_13_34-nomachine - nx07

Here is a preview of the tracker/piston setup: 
https://github.com/blackcathj/macros/tree/FGEM_fsPHENIX_DualECAL

johnlajoie:
Hi @blackcathj - I have just committed a change to the mapping file that should now leave room for the magnet piston: 
new_mpc_gap

This now only uses 278 of the MPC crystals.  We could use more but I would need to redo the PbGl supermodule mapping, and Sasha's recent conversations with Akio seem to push us in the direction oft using the PbSc instead of the PbGl, so I'd like to let this settle out before I redo the mapping again.   What we have for now is certainly valid for getting us started.

blackcathj:
Hi, @johnlajoie 

Thanks for the update, there is no more confliction with the piston anymore. 

However, it appears one additional row on (-x) and (-y) was removed. I would naively guess it is something related to the following lines in ```makeMapping_FEMC_fsPHENIX.C```. Shall we add this row before merging everything? 

```c++
	      if( ((x_corner_1<femc_xmin) && (y_corner_1<femc_ymin)) ||
		  ((x_corner_2<femc_xmin) && (y_corner_2<femc_ymin)) || 
		  ((x_corner_3<femc_xmin) && (y_corner_3<femc_ymin)) || 
		  ((x_corner_4<femc_xmin) && (y_corner_4<femc_ymin)) )
		continue; 

	      if( ((x_corner_1>femc_xmax) && (y_corner_1>femc_ymax)) ||
		  ((x_corner_2>femc_xmax) && (y_corner_2>femc_ymax)) || 
		  ((x_corner_3>femc_xmax) && (y_corner_3>femc_ymax)) || 
		  ((x_corner_4>femc_xmax) && (y_corner_4>femc_ymax)) )
		continue; 
```

Beam view of the calorimeter. The grey circle is the piston magnet which has no more overlap with FEMC (cyan). But the MPC coverage is not symmetric right now:
poweron7

johnlajoie:
Fixed in commit e1ef072 to calibrations, the new mapping file has a nicely symmetric MPC and PbGL array:

newmod

I also checked overlaps for all the towers, no overlaps found so hopefully this is good to go.

PR #18, opened 2016-04-01
Fix inner radii of the beampipes based on engineering drawing.

damcglinchey:
Change the inner radii of the beam pipes to match the engineering diagram. Both the Be and Al beam pipes should have inner radii of 2.0005 cm
[22906rB_PHENIX_(v2)_Beam_Pipe_Asm_CAP.pdf](https://github.com/sPHENIX-Collaboration/macros/files/199925/22906rB_PHENIX_.v2._Beam_Pipe_Asm_CAP.pdf)


pinkenburg:
Yes the drawing definitely says 2.0005 cm radius

PR #20, opened 2016-04-13
new G4Bbc library

mccumbermike:
BBC code base moves to g4bbc from g4vertex.


blackcathj:
Please use this macro with today's new build: 
/afs/rhic.bnl.gov/sphenix/sys/x8664_sl6/new.2

PR #21, opened 2016-04-15
Online production macro for T1044 beam test

blackcathj:
Online production macro is now separated from the code development location. See more here: 
https://github.com/sPHENIX-Collaboration/coresoftware/pull/134


blackcathj:
Production library and calibration constant for EMCal and HCals are now in the nightly build. Merging this macro too.

PR #30, opened 2016-06-24
adapt macros to PHG4Parameter interface

pinkenburg:


pinkenburg:
The macros need updating to play with the PHG4Parameter class which now handles the parameters, rather than leaving it up to every subsystem to come up with its own implementation

PR #34, opened 2016-07-29
Add macros to run EIC Detector geometry

nfeege:
Adding macros to build full EIC detector geometry.

eicdet

Overlapcheck of all detectors yields overlaps of:

CEMC_1_sec1 with CEMC_1_sec0 volume's (message repeated for all subsequent CEMC_1_sec###) --> Is this a known overlap? Should be the same for default sPHENIX.

hEcalTower_envelope with OuterHcal volume's --> Corner of forward-ECAL clears the physical outer-HCAL, but touches the rectangular boundary box.

BH_1 with hHcalTower_envelope --> hHcalTower_envelope is cone shaped and extends beyond the physical towers, hence this overlap.

### DIFF
--- a/macros/g4simulations/Fun4All_G4_EICDetector.C
+++ b/macros/g4simulations/Fun4All_G4_EICDetector.C
@@ -68,14 +68,14 @@ int Fun4All_G4_EICDetector(
 
   // EICDetector geometry - barrel
 
-  bool do_DIRC = true;
+  bool do_DIRC = false;
 
   // EICDetector geometry - 'hadron' direction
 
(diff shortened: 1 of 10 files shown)


nfeege:
Updated geometry:

screen shot 2016-08-05 at 11 48 43 am

Overlaps of CEMC sectors and BH with forward HCAL resolved.

Remaining overlaps: forward-ECAL (FEMC) and Aerogel with 'empty' corners of barrel HCAL. Need to adjust envelope for barrel HCAL to fix this.

PR #38, opened 2016-11-08
more fine grained switches for debugging, better positioning of hi eta prototype

pinkenburg:
The prototype macros turned into long hard to read spaghetti macros. In order to make them more usable I introduced switch settings which enable us to run selected parts of the reco chain. The switches for each step add in the settings of the previous module switch to prevent running of useless combinations.
The hi eta prototype locations are better adjusted now. It's still not perfect but we can probably go with that. The surrounding black hole is not adjusted yet


pinkenburg:
This does not interfere with any other of our projects, but it does give a better representation of the location and beam angle for the 3rd prototype

PR #40, opened 2016-11-14
set number of scintillators per tower to 4 in inner hcal

pinkenburg:
The number of scintillators per tower is now a parameter which then can be picked up by the cell reconstruction (still needs to be implemented there). Also the particle gun was added to the main macro as option and the subsystem flags now incorporate the flag settings of the input modules they need.


pinkenburg:
need this to patch up issue https://github.com/sPHENIX-Collaboration/coresoftware/issues/209 for now.

PR #49, opened 2017-02-22
updated with optimized number of required tracking layers

adfrawley:
Update of G4_Svts_maps+IT+tpc.C to allow tracks with 4 missed layers instead of 30 missed layers. The 30 missed layers was screwing up the tracker cluster association somehow. 

After this update, the tracking efficiency for 2-pion events is 90% and the cluster distributions look reasonable for reconstructed tracks, see attached plots.
[clusters_per_track_by_layer_4missed.pdf](https://github.com/sPHENIX-Collaboration/macros/files/794799/clusters_per_track_by_layer_4missed.pdf)


blackcathj:
Thanks for the tuning, Tony!

PR #54, opened 2017-05-18
Si ladders new tpc

adfrawley:
I removed a bunch of obsolete tracking setup macros, and I added a new macro "G4_Svtx_maps_ladders+intt_ladders+tpc_KalmanPatRec.C" that sets up the ladders for the MVTX and INTT, and uses the latest TPC setup (with 60 layers, as suggested by Carlos). It uses Carlos new cluster position smearing - set to give 125 micron TPC cluster resolution. It has the option of using Haiwang's KalmanPatRec code (not yet in the repository) or Haiwang's TruthPatRec code (exists in the repository now, and is selected as default). 

The TPC cluster smearing code will be available in the next nightly build, until then you have to build g4detectors for this new tracking setup to run.

The Kalman refitter currently throws many exceptions and produces somewhat poorer momentum resolution when used with the TruthPatRec option than it does with the the KalmanPatRec option. Haiwang is looking into this now. The full KalmanPatRec code will hopefully be in the repository next week.

If you don't want to run the track refitting, you can set "#define _ONLY_SEEDING_".


blackcathj:
Hi @adfrawley , 

Thank you for the new tracking macro and clean up! 

I added few minor edits to put the print out behind verbosity checks, so user do not see tracker construction messages. 

Problems observed but not directly affect merging of this pull request:
* There is also a few warning from Geant4 GDML import. It does not affect running, but good for us to clean up in the future (I have an idea). 
* There is a chance that the Kalman filter crashes during fitting stage. As I understand, @HaiwangYu is working on it.

Suggest merge and start test productions, ATTN: @pinkenburg

pinkenburg:
works, used it for testing of the new cellreco code - but lots of error messages from the tracking

PR #60, opened 2017-05-27
latest initialization of KalmanPatRec

adfrawley:
This is the macro that was used to generate the performance plots shown in the tracking meeting today.
It has Haiwang's latest initialization of the KalmanPatRec code - it just uses all the defaults.
It runs with the latest new libraries.

### DIFF
--- a/macros/g4simulations/G4_Svtx_maps_ladders+intt_ladders+tpc_KalmanPatRec.C
+++ b/macros/g4simulations/G4_Svtx_maps_ladders+intt_ladders+tpc_KalmanPatRec.C
@@ -1,8 +1,5 @@
 #include <vector>
 
-#define _USE_NEW_TPC_
-//#define _ONLY_SEEDING_
-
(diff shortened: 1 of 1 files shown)


blackcathj:
Hi @adfrawley and @HaiwangYu  , 

Running few tests of this new tracking macro and nightly-build lib with single and 50-electron events, it appears very few track was reconstructed. In the same setting, the current default macro without this pull request reconstruct almost all the tracks with the same test (although has lots of error print). 

Could you double check that?

adfrawley:
Hmm. OK, I will try to understand that. Note that the macro in the repository does not run the KalmanPatRec stuff, it just runs truth pattern recognition followed by tracking.


The reason I committed this new macro is that the setup for the KalmanPatRec code needs to be different from what is in the repository for it to work - presently only truth pattern recognition works for the repository version.


Can you point me to the area that you tested it in?


Thanks

Tony

PR #66, opened 2017-06-19
Tracking cleanup june19

adfrawley:
Changed G4Setup_sPHENIX.C to use the 40 layer TPC tracking setup.
Added a switch to enable the primary track vertex to be included in the track fit (defaults to false).
Lowered threshold setting on maps detector pixels to 0.1 mips from 0.25 mips.

### DIFF
--- a/macros/g4simulations/G4Setup_sPHENIX.C
+++ b/macros/g4simulations/G4Setup_sPHENIX.C
@@ -20,8 +20,7 @@ void G4Init(bool do_svtx = true,
     }  
   if (do_svtx)
     {
-      gROOT->LoadMacro("G4_Svtx_maps_ladders+intt_ladders+tpc_KalmanPatRec.C");  // default tracker setup with laddered silicon trackers and 60-layer TPC (approximation of final design)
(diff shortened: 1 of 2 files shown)


blackcathj:
Hi @adfrawley  and @pinkenburg 

Thanks Tony for the update! For the changes in ```G4Setup_sPHENIX.C```, shall we keep the previous 60-layer TPC option as default until we could refresh the calorimeter simulation production data set? Like before, a commented-out 40-layer option in ```G4Setup_sPHENIX.C``` still allows tracking expert to run 40-layer TPC tests.

adfrawley:
Hello Jin,


Ah, I forgot about that. The problem is that all new tracking sims should be run with the 40 layer setup, so real confusion is possible. Maybe we could set a flag based on whether readhits is true?


Thanks

Tony


pinkenburg:
Instead of using two separate 500 line macros which differ in only a few lines the macros should be merged and we pass down the number of layers from the top macro (the number of layers then serves to distinguish the cases). Then the 60 layer code can be phased out easier (we don't have to tell everyone to change their setup to use yet another macro name)

adfrawley:
Hi Chris,


Good point. I will move the 60 layer TPC as an option into this macro.


Thanks

Tony

PR #76, opened 2017-08-29
Test switch for inner Hcal absorber material to Aluminum

sen1:
Change both instances of "material_Al" to true inside G4_HcalIn_ref.C to update your absorber material to "G4_Al". Default is still SS310.


blackcathj:
Thanks, Abhisek! 

Please set ```const bool inner_hcal_material_Al = true``` in ```macros/g4simulations/G4_HcalIn_ref.C``` in order to simulate inner HCal absorber with Aluminum. This macro need to be set consistently for simulation production and analysis stages.

PR #89, opened 2017-11-30
Updates pad segmentation (T Frawley).  Updates gas character (T. Hemmick).

tkhemmick:
These updated macros DO implement the current pad segmentation (Tony Frawley).  They also update the gas characteristics in a manner selectable through enumerated self-consistent sets.  It also includes a "ByHand" set to allow users to play what-if scenarios (Tom Hemmick).

These macros REQUIRE the prior TKH pull request from core in order to run.

What is not (yet) done is optimization of the pad aspect ratio to minimize overall occupancy.  Stay tuned...


blackcathj:
Merged into #89

PR #96, opened 2018-01-23
Add field return

blackcathj:
Jamie reminded that the flux return plug door of sPHENIX is not implemented in the default sPHENIX simulation (a thinned version is in fsPHENIX and ePHENIX simulation). Although flux return is outside the active acceptance of sPHENIX, recent tracking study does cautious us against albedo background originated from back scattering of heavy materials outside the tracking volume. 

This pull request introduce the plug door to the default sPHENIX simulation. Here is two detector display. The geometry of the plug door come from a preliminary engineering drawing which is about 30 cm thick. 

In engineering drawing:
image

In simulation:
image
image

Checking its effect on TPC tracking background. Since the plug door is far away from the TPC, the change in TPC clsuter multiplicity is minimal. Identical generated collisions and vertex location are used in these comparison, while the Geant4 cycles are independently randomized, which lead to some fluctuation in the difference. 

image

Although the change in background is small, still introduce the plug door as an option of the simulation for completeness. The pair of doors are OFF by default, since they does not change tracker backgroudn much and it prolongs the Geant4 simulation by 60% of CPU usage on hadronic showers inside the plug door.

### DIFF
--- a/macros/g4simulations/Fun4All_G4_sPHENIX.C
+++ b/macros/g4simulations/Fun4All_G4_sPHENIX.C
@@ -82,6 +82,9 @@ int Fun4All_G4_sPHENIX(
   bool do_hcalout_cluster = do_hcalout_twr && true;
   bool do_hcalout_eval = do_hcalout_cluster && true;
 
+  //! forward flux return plug door. Out of acceptance and off by default.
+  bool do_plugdoor = false;
+
   bool do_global = true;
   bool do_global_fastsim = true;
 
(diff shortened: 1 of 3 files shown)


blackcathj:
From Jan 23 meeting, Chris' comment: making flux door inactive by default even when user choose to include it in simulation. The consideration is active flux door records large amount of hits to DST tree during the simulation. It will break memory limit in central AuAu simulations

PR #102, opened 2018-03-19
Calorimeter beam test macro : add leading order HCal energy scale calibration

blackcathj:
Calorimeter beam test macro update: add leading order HCal energy scale calibration from Xu Sun (GSU). Xu's slides for details: [hcal_beamtest_energycalibration.pdf](https://github.com/sPHENIX-Collaboration/macros/files/1825603/hcal_beamtest_energycalibration.pdf)

### DIFF
--- a/macros/prototype4/Fun4All_TestBeam.C
+++ b/macros/prototype4/Fun4All_TestBeam.C
@@ -100,13 +100,13 @@ void Fun4All_TestBeam(int nEvents = 1000,
   {
     // leading order energy scale from Xu Sun
     const double cin_cali = 0.00297968;
-    const double cout_cali = 0.000215485
+    const double cout_cali = 0.000215485;
 
-        calib = new CaloCalibration("HCALIN");
+    calib = new CaloCalibration("HCALIN");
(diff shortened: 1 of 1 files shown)


sPHENIX-GitHub-l:
Interesting and almost impossible to understand - x15 differerrence in calibration scales. One would expect x2 from light collection dependence on the length of embedded fiber times x1.5 due to sampling fraction difference in the older HCals. To go from x3 to x15 one needs to be very creative with ???????? - tail cutting? No quick answer except saying that cuts based upon expecting to see Gaussian in individual detector component are likely to be grossly flawed. 
Sorry for the inconclusive comment but we need to discuss this picture with more details available.
Edward

blackcathj:
Good point Edward. The problem is traced down to the unpacking software swapped low-gain and high-gain outer HCal channels, and therefore this pull request used the high gain in the outer HCal. 

Both problems are being resolved in https://github.com/sPHENIX-Collaboration/macros/pull/104  and https://github.com/sPHENIX-Collaboration/coresoftware/pull/423

PR #103, opened 2018-03-22
use tower-by-tower calibration for EMCal

blackcathj:
More information in https://github.com/sPHENIX-Collaboration/calibrations/pull/31


blackcathj:
sPHENIX-Collaboration/calibrations#31 is now in the lastest nightly build. Please use a fresh login of RCF to use the tower-by-tower EMCal calibration with this macro.

PR #106, opened 2018-04-25
mRICH

sunxuhit:
Implement mRICH detector into G4_Aerogel.C.


blackcathj:
Hi, @nfeege : could you coordinate with @sunxuhit on the integration of mRICH geometry with the EIC/fsPHENIX simulations? The current geometry map covers all the way down to eta = 1. We probably need to shrink it according to the new mRICH envelop.

blackcathj:
@sunxuhit , https://github.com/sPHENIX-Collaboration/macros/pull/107 is merged. 
Could you and @nfeege  double check the mRICH fit again? It will need some adjustment to resolve the confliction on macro setup too. Thanks!

PR #165, opened 2019-03-19
copy G4_Tracking.C from ModularizedTrackReco branch

pinkenburg:
the G4_Tracking.C macro implements the modularized tracking. The other macros in that branch were outdated which made merging difficult. So I just copied the only relevant macro over


blackcathj:
It appears with many track run, it is likely to get stuck in a loop and never complete the tracking step. Reverting.

PR #173, opened 2019-04-26
Modifications needed after new MVTX detector group module

ycorrales:
This PR introduces changes in G4_Tracking following up new maps subsystem implementation (coresoftware PR #626)


blackcathj:
Thanks for the update. 

Could you update a similar setup under the EIC macro too? That is at `Svtx()` in https://github.com/sPHENIX-Collaboration/macros/blob/master/macros/g4simulations/G4_Svtx_maps_ladders%2Bintt_ladders%2Btpc_KalmanPatRec_EIC.C
which runs with `Fun4All_G4_EICDetector.C`

PR #184, opened 2019-06-03
Fix InttHitReco module name

pinkenburg:
The parameter is supposed to be the module name, not the detector so one can see in printouts which module is running. The detector name is now initialized in the code, it's still hokey and will break if you rename the detector.


pinkenburg:
The new code is merged and has been build in the meantime.

PR #202, opened 2019-09-10
Updated w/ fix for the IHCal material switch between Al and SS310; al…

cfmcginn:
…so tweaked the usegun option so defaults avoid weird response behavior. This PR is as discussed in sPHENIX SW meeting on 2019.09.10; Slide are available here:

https://indico.bnl.gov/event/5481/contributions/31285/attachments/24515/36296/sPhenixCaloZSSim_20190910.pdf

### DIFF
--- a/macros/g4simulations/Fun4All_G4_sPHENIX.C
+++ b/macros/g4simulations/Fun4All_G4_sPHENIX.C
@@ -1,4 +1,8 @@
 #if ROOT_VERSION_CODE >= ROOT_VERSION(6,00,0)
+
+//cpp
+#include <string>
+
 #include <fun4all/SubsysReco.h>
 #include <fun4all/Fun4AllServer.h>
 #include <fun4all/Fun4AllInputManager.h>
@@ -27,6 +31,7 @@
 #include "G4_HIJetReco.C"
 #include "G4_DSTReader.C"
 #include "DisplayOn.C"
+
 R__LOAD_LIBRARY(libfun4all.so)
 R__LOAD_LIBRARY(libg4testbench.so)
 R__LOAD_LIBRARY(libphhepmc.so)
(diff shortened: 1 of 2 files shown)


cfmcginn:
@blackcathj @pinkenburg - hopefully you automatically receive notification but just in case

pinkenburg:
The change in Fun4All_G4_sPHENIX.C will break the particle gun. We might want to change this in the future but the current setup uses an eta range of -0.01 to 0.01 using the particle generator (the registration of the particle gun is commented out) I don't think 0.01 always hits the gap at eta=0 so this should produce okay non confusing results

cfmcginn:
Hi @pinkenburg sounds good; just for my own studies, what about the change to Fun4All would break particle gun workflow? (Using this to refer to usegun=true, you are right, particle gun itself is commented out).

pinkenburg:
Hi Chris,
you had just removed the particle gun (PHG4ParticleGun) which is anyway not running by default (the registration with Fun4All is commented out, that must have happened sometime in the past and we just got used to having to modify this when using the gun). What is being used is our single particle generator PHG4ParticleGenerator. In the original macro it does have a range in eta, if the range is too small to send particles outside of the midrapidity gap we should change it. Otherwise it actually works as intended. We do not want to change this into a particle gun by restricting the range to a single value. In this case you should use PHG4ParticleGun. Personally I prefer to set the values in the generator methods:
pgen->set_name("geantino");
rather than
std::string particleName = "geantino";
...
...
..
...
...
pgen->set_name(particleName);

that makes this just easier to read and follow. Naturally if you have to do some math with them or reuse them later, it's a different story. Error conditions should be handled by a gSystem->Exit(1), a return 1 will just make the macro return to the command line. In this case it would be obvious but I don't want to deal with partially working macros which are executed by other macros confusing everyone (and nobody sees error messages which are somewhere in the middle of the usual blabbering).

cfmcginn:
Hi,
Thanks, in future I will use gSystem->Exit(1) over return 1 for errors. As to the default position of the gun in eta, I would argue based on these plots:

https://www.dropbox.com/s/2o9dfb3hdo4j7j7/20190911_20190911.pdf?dl=0

that central value of zero still gives confusing response results for particles that ought be dominated by CEMC (layer 3) to the newcomer over my proposed eta value centered on a mid-rapidity ecal crystal, even with the 0.01 window (limited stats but I think effect is clear comparing the two eta windows; note that even tho the centered zero window says -0.009 to 0.009 thats just an artifact of my little string code and the true cut really is -0.01 to 0.01).

Expanding the window is I suppose a solution here but you start to lose the gun-like behavior as the window gets much larger. I don't mind leaving current code as is but maybe it would be appropriate if I add a few comments to help folks pick an eta/understand the default behavior?

Thanks again for your help!

PR #213, opened 2019-10-23
Fixed FEMC Separation Bug

jdosbo:
Jin pointed out a bug with running the EIC framework in root5 after the FEMC fsPHENIX/EIC separation. 

Tested with both root5 and root6 on rcas - bug of not working with root5 appears to be fixed.


blackcathj:
Appears work well. Thanks @jdosbo 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-pipeline/3970/console

PR #238, opened 2020-05-04
New macro scheme

pinkenburg:
This PR is a proposal how to handle our out of control macro handling uding Fun4All_G4_EICDetecotr.C as example. It replaces those do_XXX switches which need to be propagated as arguments to the G4Init() and G4Setup functions by an Enable namespace which can be accessed in the G4_<subsystem> macros making it easy to add new ones.

GlobalVariables.C now holds all variables

The size of the black hole is adjusted according to the selected subsystems (which therefore have to be aware of their own dimensions). One can make it visible for easier debugging. The black hole is now adjacent to the outermost volume(s) - the distance can be chosen by setting BlackHoleGeometry::gap in the main macro.

If the size of the detectors exceed the default size of the world, the world volume will be adjusted accordingly (with 1m space added). It looks like our eic detector might have exceeded this already.

The changes in the macros do not have any effect on the simulations (for direct comparison the black hole and the resizing of the world (in G4Setup_EICDetector.C) needs to be disabled in the new sims and the black hole commented out in the old sims since these are different which then affects the random number generartors


macros/g4simulations/GlobalVariables.C
```diff
@@ -1,5 +1,32 @@
...
+  static bool BLACKHOLE = false;
+  static bool CEMC = false;
+  static bool DIRC = false;
+  static bool EEMC = false;
+  static bool FEMC = false;
+  static bool FHCAL = false;
+  static bool HCALIN = false;
+  static bool HCALOUT = false;
+  static bool MAGNET = false;
+  static bool OVERLAPCHECK = false;
```
blackcathj, line 17: Is this boolean used? It appears the macros still calls `static bool overlapcheck = false` on line 2

macros/g4simulations/G4_BlackHole.C
```diff
@@ -0,0 +1,81 @@
...
+  blackhole->set_double_param("length", blackhole_length); // make it cover the world in length
+  blackhole->set_double_param("place_z",blackhole_zpos);
+  blackhole->set_double_param("thickness",BlackHoleGeometry::gap/2.); // it needs some thickness
+  if (BlackHoleGeometry::visible)
+  {
+    blackhole->set_color(1,0,0,0.7);
+  }
+  blackhole->BlackHole();
+  blackhole->SetActive(); // always see what leaks out
+  blackhole->OverlapCheck(true);
```
blackcathj, line 34: Less verbose in default running with following?
```
blackhole->OverlapCheck(overlap);
```
Similar for the next three instances

macros/g4simulations/Fun4All_G4_EICDetector.C
```diff
@@ -192,7 +196,7 @@ int Fun4All_G4_EICDetector(
 
   // establish the geometry and reconstruction setup
   gROOT->LoadMacro("G4Setup_EICDetector.C");
-  G4Init(do_tracking,do_cemc,do_hcalin,do_magnet,do_hcalout,do_pipe,do_plugdoor,do_FEMC,do_FHCAL,do_EEMC,do_DIRC,do_RICH,do_Aerogel);
+  G4Init();
```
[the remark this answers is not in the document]
pinkenburg: Hi Jin,

it's still very rough to show were I think things should go (you caught me having checked in what I was working with - that PR is testing the waters and should definitely not be merged in the current form). Everything is open for discussion.
I took the EIC macro since there is a lot of interest in flexibility (and I won't break sPHENIX) but this should also be applied to sPHENIX then. The underlying subsystem macros are mostly ready.

I did not want to break existing privately changed macros (that's why the overlap is still there). Those macros will not get the close fit black hole or increase in the world volume but should stay functional - at least for a while - to make this less disruptive. 

The main problem will be the tracking macro, we have been very lax with global variables. Since all macros are #included the scope of every variable declared outside of the functions extends to the whole project. That's why they should move into distinct namespaces.

We can move this to the DST reader but I am not sure what to do about local flags (like do_cemc_cell) which stay local in the Fun4All macro. Replacing those with an enormous list of booleans in the Enable namespace sounds counter productive. Given that new developers are coming I want to avoid single large files which everyone has to modify. Probably separate namespaces loaded with the G4 macro is the way to go? That would also take care of subsystem specific flags like the eic beam pipe (this would blow up if you load the wrong macro but I would see this as a failsafe).

The overlap check for the blackhole is essential - you can easily place this inside your detector and you'll never figure out why you are loosing hits (yes that happened to me which made me think about this). I'd rather look into suppressing the messages - even if it takes a G4 hack (same for the gdml, these messages are hardcoded - not sure who came up with that).

pinkenburg:
The Enable namespace is now distributed over the G4_<subsystem>.C macros. This avoids a central place where things collide. This PR only changes the EIC detector setup, the changes in the common macros do not affect the old Fun4All_G4_sPHENIX.C and Fun4All_G4_fsPHENIX.C implementations. We can play with the EIC until we are happy with the new scheme

PR #240, opened 2020-05-06
added macro to plot number of clusters per layer per track.

hupereir:
This PR adds the macro for plotting ncluster/track/layer as also discussed in https://github.com/sPHENIX-Collaboration/coresoftware/pull/805


blackcathj:
New plots from QA tests (w/o reference)
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/176/QA_20Report/

PR #246, opened 2020-06-12
Add LANL FST_5 in DST tree

pingwong:
LANL FST_5 was missing in G4_DSTReader_EICDetector.C. Now, it is back in the macro.

To admin, I didn't plan to create a pull request for all the changes I made in this directory. If you can, please accept the changes on the G4_DSTReader_EICDetector.C (commit c6fd072) alone.


blackcathj:
Thanks @pingwong ! I need to change it a bit so the default behavior is unchanged for other detectors.

PR #247, opened 2020-06-14
added micromegas subsystem and reco modules

hupereir:
This PR adds micromegas and reconstruction modules to G4_Tracking.C
By default micromegas are disabled. To enable, change enable_micromegas from false to true either in G4_Tracking.C or in the calling macro

### DIFF
--- a/macros/g4simulations/G4_Tracking.C
+++ b/macros/g4simulations/G4_Tracking.C
@@ -14,6 +14,10 @@
 
 #include <g4main/PHG4Reco.h>
 
+#include <g4micromegas/PHG4MicromegasDigitizer.h>
+#include <g4micromegas/PHG4MicromegasHitReco.h>
+#include <g4micromegas/PHG4MicromegasSubsystem.h>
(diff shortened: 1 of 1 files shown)


macros/g4simulations/G4_Tracking.C
```diff
@@ -82,6 +89,10 @@ int n_tpc_layer_mid = 16;
 int n_tpc_layer_outer = 16;
 int n_gas_layer = n_tpc_layer_inner + n_tpc_layer_mid + n_tpc_layer_outer;
 
+///////////////// Micromegas
+bool enable_micromegas = false;
+const int n_micromegas_layer = 2;
```
blackcathj, line 94: micromegas would contribute to the tracking and the momentum resolution. Shall we default it as `n_micromegas_layer = 0`, so the default behavior do not change?

blackcathj: Ignore my question. Missed the line above it....

PR #249, opened 2020-06-16
Lanl fst

pingwong:
(1) Separate FST from the FGEM macro
(2) Add a new barrel tracker with 5 layers. Each layer has 7 sub layers.
(3) Both barrel tracker and FST are from by Xuan Li's design.
Screen Shot 2020-06-16 at 4 12 47 PM


pingwong:
> Please do not merge, I need to coordinate this with my changes

Do I need to un-merge it? I don't really know what to do, so I am not going to anything right now. I do fix two bugs in the G4_Barrel_EIC.C though.

PR #251, opened 2020-06-23
making PHTpcTracker default seeding algo

bogui56:


blackcathj:
After settling down with the low level code checking, here is quick summary of the tracking QA changes in the last nightly check of the master branch coresoftware and macros: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/sPHENIX_CoreSoftware_MasterBranch/882/

# Observations

1. Significantly more TPC hits are picked up with this new seeding, with improvement on the unique-matching track reconstructed efficiency in the low pT region (only require unique best matching truth-reco pair, but not requiring momentum matching).  
2. This comes at a significant loss of inner silicon tracker matching efficiency
3. In the high occupancy cases, resolution is worse and more tails are reconstructed away from pT reco/truth~1, which leads to more failed Gauss fitting in momentum resolution calculation.  

Both 2 and 3 suggest more noise or merged TPC hits are picked up by the new seeding, which may require tuning of seeding and/or a 2nd stage clean up in fitting to reject biasing clusters from the track. 

# Snapshot of the QA plots

Copying them out from the QA report for quick reference in the future

## Low occupancy QA
image
image
image
image

## High occupancy QA
image
image
image
image

# Next

Suggest reevaluate whether we would like make this seeding switch or further adjustment in the next Tue software meeting. 
@bogui56

blackcathj:
And another seeding related memory leak at `coresoftware/offline/packages/PHTpcTracker/externals/kdfinder.hpp:3160`: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-valgrind-pipeline/1811/valgrindResult/pid=166717,0x3fff0/
which failed the last master branch check

PR #254, opened 2020-07-02
Sphenix new macros

pinkenburg:
This PR puts the sPHENIX macros into the new scheme. The barrel tracking of sPHENIX and fsPHENIX was merged into one again. One should now be able to switch between seeding algorithms in the Fun4All macro rather than editing the G4_Tracking.C directly. In general options can easily be propagated between macros using namespaces, no more painful funneling via function parameters. The TPC drift velocity for the pileup generator is hardcoded in G4_Input.C (as well as in 2 TPC modules) - we need to find a solution for this


pinkenburg:
Please have a look at this, given the collaboration meeting next week this PR will only be merged after the meeting. But the upcoming hijing production will use this (there might be more changes to make the productions easier to run)

hupereir:
On the evaluator crash:

    yeah I did not test if things are running with the evaluator
    most likely what happens is that the evaluator finds g4hits keys that cannot be associated to any container (because it does not know about the container for micromegas, does not load the node, etc.)
    I had the same problem in trackprop before.
    I will provide a fix later today.

Hugo

hupereir:
> Please have a look at this, given the collaboration meeting next week this PR will only be merged after the meeting. But the upcoming hijing production will use this (there might be more changes to make the productions easier to run)

Hi Chris, 
I will test this either later today, or tomorrow the latest. 
Hugo

hupereir:
Hello Chris,
Looking at this patch (and testing), now.
One tiny (but possibly annoying) request ? G4MICROMEGA -> G4MICROMEGAS, and G4_Micromega.C -> G4_Micromegas.C
This would be consistent with the way the detector is named everywhere else, and is really the name of the detector ...  (the S is not for plural, but for GAS) 

Hugo

hupereir:
Comment on G4_Input_Simple
Hi again, 
so playing around with the new macro system (and updating my local macros accordingly), something that I find cumbersome is that G4_Input_Simple essentially duplicates the API of PHG4SimpleEventGenerator in order to provides ways to customize it to the main macro. 
Now, this is cumbersome because it is not enough for me to know the API of PHG4SimpleEventGenerator to make local modifications, I must also know the one of the macro wrapping around it. And if one makes changes to the generator API (I did ! to accomodate using realistic input pT distribution), one has to essentially duplicate it to the macro. 
Maybe it would be easier to remove the duplicated API, set "sane" parameters in InputSimpleInit, and let InputSimpleInit return a pointer to the subsysreco so that people can modify it directly in the calling macro ? This would allow to keep the Input macro clean, but keep the flexibility in the main macro, without having to duplicate the API. 
Just for discussion ...

hupereir:
Aside from the comments above, it is all working well, and I could reproduce results obtained (for micromegas) with the old set of macros. No objection against using those for the Hijing production.

pinkenburg:
Hi Hugo,
you definitely have a point with the generator parameters. It was a bit of a chicken and egg issue but a 2 step process took care of that. In the first step the enabled generators are created and put in a namespace which is accessible in the main macro. One can now talk directly use the generators API in the main macro. In a second step they are registered. That removes the duplication of the APIs which would have been a permanent catch-up race.
Also MicroMega --> MicroMegas. I just saw it should be Micromegas, I'll change this just now

hupereir:
Hello @pingwong 
I just checked the latest patch. I think indeed the handling of the input generators is much more convenient and flexible now. Thanks !the m
One more (last?) nitpick, which you can feel free to ignore:  the method names inside G4_Micromegas should also be Micromegas rather than MicroMegas for consistency with elsewhere. 
Other than that, no objection from my side against pushing this. This looks much cleaner than before.

pinkenburg:
The merge with the master shot this one to pieces. We should just never do this. I'll close this and open another one

To github.com:pinkenburg/macros.git
 ! [rejected]        sphenix-new-macros -> sphenix-new-macros (non-fast-forward)
error: failed to push some refs to 'git@github.com:pinkenburg/macros.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

with git pull:

pinkenbu@rcas2601 g4simulations]$ git pull
X11 forwarding request failed on channel 0
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

If you wish to set tracking information for this branch you can do so with:

    git branch --set-upstream-to=<remote>/<branch> sphenix-new-macros

pinkenburg:
Closing this

PR #265, opened 2020-07-15
Plotting code for tracking purity and tracking efficiency histograms

rosstom2232:
Updated QA_Draw_ALL.sh and QA_Draw_Tracking_TruthMatchingOverview.C to be able to plot new histograms and added new file QA_Draw_RecoTruthMatching.C to create tracking purity plot


blackcathj:
Thanks! @rosstom2232 . Here is the test run results for track purity: 

Low occupancy: 
image


High occupancy: 
image


Please note in both cases, flat generated pT spectrum between [0-2] GeV and [2-50] GeV ranges are used. Therefore, the purity plot do not represent the purity for a physical collision situation, and it is just for tracking quality monitoring. 

Thomas also added the eta-dependence of efficiency to preview pT-dependence-only plot: 
image

PR #267, opened 2020-07-17
Acts Introduction

osbornjd:
This PR introduces Acts into the main macros repository. It is not used by default since it is in a regularly changing state. Thus, this PR should not actually change anything in the default macro running. The reason for introducing it now is that it would be useful to be able to run the tracking QA on the Acts fitting to see how changes in `coresoftware` affect basic observables like the pT and DCA resolution. 

One comment for now - the Acts fitting seg faults when used with the `PHTpcTracker` seeder. A quick look of mine showed the seg fault happens [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/61cc6697d7debf577dd386bbe8d6ddd9c90a16d2/offline/packages/trackreco/PHActsTracks.cc#L101), because the `SvtxVertex` object retrieved is a `nullptr`. I'm guessing that this means somehow the `PHTpcTracker` is overwriting and/or updating the results of `PHInitVertexing` incorrectly. I didn't dig deeply into this yet, but figured it would be worthwhile to let @bogui56 know.

Thus, for those that want to run Acts, it should be run with the `PHHoughSeeding` seeder (for now).


blackcathj:
Hmm, Jenkins think this will break the macros under `new` / default environment: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-pipeline/8192/console 

@pinkenburg , is there a compiler checker we can put around the header files to avoid loading the ACTS modules unless it is gcc8?

osbornjd:
I found that there is a predefined GNU macro `__GNUC__` which is supposed to tell you the version of gcc, but for some reason it tells me that the version is 4.2.1 in both the default environment and the Acts environment (which has gcc 8.3, so I'm not sure why it gives 4.2.1...)

blackcathj:
valgrind error comes from `PHGenFitTrkProp` and not related to this pull request: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-valgrind-pipeline/1875/valgrindResult/pid=190338,0x311eb/

osbornjd:
The CI will fail because of the ongoing Cling/Boost issue. Chris is planning on patching our version of root, after which we can restart the CI. This PR should otherwise be good to go for both the new and old macro scheme, pending the CI success after the root patch.

PR #279, opened 2020-08-11
Fit curves added to to upsilon overview plots

rosstom2232:


blackcathj:
New plot from the check runs: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/466/QA_20Report/

image

PR #281, opened 2020-08-13
Variable CEMC parameters from calibrations file

griffsk:
Pull request for variable zero suppression, pedestal, and GeV per ADC in CEMC. This request implements the changes made in PR #889 in coresoftware and PR #64 in calibrations. Previous structure with statically defined pedestal, zero suppression threshold, and GeV per ADC remains but is commented out. This may be useful for quickly testing new parameters since it does not require generating a new calibrations xml.


blackcathj:
Check completed: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master/2980/

The change in the EMCal tower distribution is subtle but visible: 
image

PR #291, opened 2020-09-08
Adding gaussian fits with sigma labels to appropriate QA plots, also …

rosstom2232:
…adding new Vertex resolution plots with respect to number of tracks reconstructed for a specific vertex.


blackcathj:
The label worked well and a very useful addition. Thanks! 

image
image

PR #317, opened 2020-10-21
added missing include

hupereir:
G4MAGNET namespace is used later in the macro. Without the include this might crash the main macro if includes are not added there in the proper order.

### DIFF
--- a/common/G4_Tracking.C
+++ b/common/G4_Tracking.C
@@ -3,6 +3,7 @@
 
 #include "GlobalVariables.C"
 
+#include "G4_Magnet.C"
 #include "G4_Intt.C"
 #include "G4_Micromegas.C"
 #include "G4_Mvtx.C"


pinkenburg:
This is problematic. I know that this macro needs the magnet but we have multiple magnet macros. Typically the magnet macro is included in the G4Setup macros where the specific magnet to be used is included which then works with the G4_Tracking.C. If we now include this in the G4_Tracking macro we will get hard to debug mixtures or the macro will crash for other implementations

hupereir:
Hi Chris, 
I agree with the issue that this raise, but on the other hand having include files that are "incomplete" and require some prior include in the calling file is bound to explode in the future. (imagine that in the next iteration some G4Magnet file require another "branched" include that cannot be added directly either, and that along the chain you are back to requiring one of the dowstream file, etc.
To me: this is a serious flaw and one should think about a more robust solution ... 
In the meanwhile I will add the missing header in my own calling macros ...

PR #333, opened 2020-11-05
Tpc distortion correction

hupereir:
This PR adds the TpcSpaceChargeCorrection module to the GenFit tracking chain. 
For now, the module will make Acts tracking fail because of the inability to assign clusters to the right surface once they have been moved in "r". So the G4_Tracking automatically disable the module if not running GenFit chain.
Also: fixed missing include in G4_Bbc.C

Question: 
Default filenames for G4TPC::distortion_filename and G4TPC::reconstruction_filename are local name:
  std::string distortion_filename = "fluct_average.rev3.1side.3d.file0.h_negz.real_B1.4_E-400.0.ross_phi1_sphenix_phislice_lookup_r26xp40xz40.distortion_map.hist.root";
  std::string correction_filename = "fluct_average.rev3.1side.3d.file0.h_negz.real_B1.4_E-400.0.ross_phi1_sphenix_phislice_lookup_r26xp40xz40.distortion_map.hist.root";
Should we not
- move the corresponding file to an official place (e.g. in $CALIBRATIONROOT)
- use the full path name in the macro (does full path understand env. variables ?), or make sure that the code looks in the official path if the file is not found in working directory ? 
This way, just turning on the flag would "magically" work.

### DIFF
--- a/common/G4_Bbc.C
+++ b/common/G4_Bbc.C
@@ -4,6 +4,7 @@
 #include <g4detectors/PHG4BbcSubsystem.h>
 
 #include <g4bbc/BbcVertexFastSimReco.h>
+#include <g4main/PHG4Reco.h>
 
 #include <fun4all/Fun4AllServer.h>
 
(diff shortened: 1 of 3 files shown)


pinkenburg:
Yes - we should add those files to our calibrations. We can use getenv() in the macros to use $CALIBRATIONSROOT, there are quite a few examples of that in our macros

PR #337, opened 2020-11-09
update LANL FST and barrel tracker

pingwong:
- G4_FST_EIC.C : update FST setup with 6 silicon detectors
- G4_Barrel_EIC.C : update barrel tracker setup. The barrel layers are group into two sets so that I can set different resolutions of different layers in G4_Tracking_EIC.C
- G4_Tracking_EIC.C : changes resolutions of the barrel detector as explained in the last bullet.


pinkenburg:
This PR has a merging conflict, in G4_Tracking_EIC.C this needs to be resolved before this can be merged

PR #339, opened 2020-11-15
QA setup for new macros

blackcathj:
This is the first of several steps to rebuild the standardized QA capability under the new macro setup [ #262 ]

QA modules are integrated in the common macros, part of each of the detector setup and jet/track reconstruction macros. They can be enabled/disabled in the main Fun4All macro like an eval step, which default to be enabled. Suggest enable these QA modules in the coming productions too. 

Next steps include introduce its jobs in Jenkins CI and build a new QA plot archival.

### DIFF
--- a/common/G4_CEmc_Spacal.C
+++ b/common/G4_CEmc_Spacal.C
@@ -2,6 +2,7 @@
 #define MACRO_G4CEMCSPACAL_C
 
 #include <GlobalVariables.C>
+#include <QA.C>
 
 #include <g4detectors/PHG4CylinderCellReco.h>
 #include <g4detectors/PHG4CylinderGeom_Spacalv1.h>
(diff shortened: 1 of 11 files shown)


pinkenburg (changes requested):
The handling of the embedding ids are overly complicated, the initial version just made sure they are distinct. If someone wants to override them they can be set explicitly in the input part of the Fun4All macro. This can be done generically since the highest embedding id is known one can always create unique id's. I don't think adding variables for the first embedding id for every single event generator adds any functionality.
The Fun4All macro has verbosity enabled but the random number printout disabled. The default macro should have this the other way round

blackcathj:
Home work done from the last meeting: 
1. Add RAVE vertex finder and output to dedicated node `SvtxVertexMapRave (IO,SvtxVertexMap_v1)`
2. Add a set of embedding ID to explicitly present multiple embedding IDs

PR #345, opened 2020-11-25
updated flags to handle new distortions from Henry

hupereir:
This pull request updates the flags in G4_TPC and G4_Tracking needed to accomodate latest distortion code from Henry.
The latter can deal with either static and per-event distortion maps (or both at the same time) and thus the configuration flags must support both configuration.


hupereir:
Please wait before merging, one also needs to update coresoftware

hupereir:
Should go together with https://github.com/sPHENIX-Collaboration/coresoftware/pull/1009

PR #353, opened 2020-12-23
KFParticle macro added

cdean-github:
This PR adds the KFParticle macro to perform Upsilon, D0 and Lc reco

The three reconstructions were tested and work without crashing although the Lc generator is still a work in progress. This will appear in a later PR to core software.

There are some base cuts which are realistic but might be tight for code testing so they can be loosened (or set to effectively no cut). The nTuple output is disabled by default while the DST output is enabled. Truth matching is also disabled by default, only enable it if you know there are truth variables. KFParticle will exit your run rather than skip the event if this is the case

### DIFF
--- a/detectors/sPHENIX/Fun4All_G4_sPHENIX.C
+++ b/detectors/sPHENIX/Fun4All_G4_sPHENIX.C
@@ -12,6 +12,7 @@
 #include <G4_HIJetReco.C>
 #include <G4_Input.C>
 #include <G4_Jets.C>
+#include <G4_KFParticle.C>
 #include <G4_ParticleFlow.C>
 #include <G4_Production.C>
 #include <G4_TopoClusterReco.C>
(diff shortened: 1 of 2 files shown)


detectors/sPHENIX/Fun4All_G4_sPHENIX.C
```diff
@@ -466,6 +473,17 @@ int Fun4All_G4_sPHENIX(
 
   if (Enable::USER) UserAnalysisInit();
 
+  //======================
+  // Run KFParticle on evt
+  //======================
+  Enable::KFPARTICLE = false;
```
blackcathj, line 479: Shall we put the ` Enable::*` section to top of the macro, after `Enable::PARTICLEFLOW`?

blackcathj: Also since `Enable::KFPARTICLE` was default to `false`, we could change this line to 
```
// Enable::KFPARTICLE = true;
```
to be consistent with others

pinkenburg:
Looks good and does not interfere with regular running. I'll kick off a build right after the merge so if someone updates his/her local Fun4All_G4_sPHENIX.C it will not complain about a missing G4_KFParticle.C

PR #364, opened 2021-01-26
Added field direction setter for PHHybridSeeder

petermic:


common/G4_Tracking.C
```diff
@@ -271,6 +271,7 @@ void Tracking_Reco()
     {
       std::cout << "   Using PHHybridSeeding track seeding " << std::endl;
       PHHybridSeeding* hseeder = new PHHybridSeeding("PHHybridSeeding");
+      hseeder->set_field_dir(G4MAGNET::magfield_rescale);
```
blackcathj, line 274: @petermic  
would you like to modify `PHHybridSeeding` to obtain the field setting via DST nodes? It would be safer compared with macro setting, e.g. during embedding. 

I think here is the clean way to consistently get the field sign in general: the field value at (0,0,0) which is a good approximation for near-vertex extrapolation too:
```c++
PHField* field = PHFieldUtility::GetFieldMapNode(nullptr, topNode);
  if (field )
  {
    const double point[] = {0, 0, 0, 0};  // x,y,z,t
    double Bx = 0, By = 0, Bz = 0;
    double Bfield[] = {std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN()};
    field ->GetFieldValue(point, Bfield);
    Bx = Bfield[0];
    By = Bfield[1];
    Bz = Bfield[2];
    B_approximation_invertex_region = Bz / CLHEP::tesla;
}
```
In reference to https://sphenix-collaboration.github.io/doxygen/d0/d76/PHRaveVertexing_8cc_source.html#l00106 and https://sphenix-collaboration.github.io/doxygen/d2/d22/PHTpcTracker_8cc_source.html#l00212

When the simulation field changes, this value is changed consistently, as it is the same interfaced used by GenFit2 and Geant4.

PR #372, opened 2021-02-19
added FST and Barrel tracker version according to 2009.0288

FriederikeBock:
- rewrote BARREL tracker geom setup to be done for each layer separately 
- implemented pitch, geom and geom according to arxiv:2009.0288 for both FST & BARREL setup


pingwong:
Do you mean arXiv:2009.02888?

FriederikeBock:
yes , sorry for the confusion.

PR #381, opened 2021-03-02
Applying beam parameters (e.g. crossing angle) to detector simulation macros

blackcathj:
Following https://github.com/sPHENIX-Collaboration/coresoftware/pull/1087 , this pull request apply the beam parameters by default to the detector simulation macros. This is applied toany HepMC-based generator, e.g. `Fun4AllHepMCInputManager`, `Fun4AllHepMCPileupInputManager`, `PHPythia8`, `PHPythia6`, `PHSartre`, `ReadEICFiles`

Two functions defines the parameters: 

* `Input::ApplysPHENIXBeamParameter(PHHepMCGenHelper *HepMCGen)` applies the 2mrad sPHENIX beam crossing as defined in sPH-TRG-2020-001
* `Input::ApplyEICBeamParameter(PHHepMCGenHelper *HepMCGen)` applies the 25mrad beam crossing and beam parameters as defined in EIC CDR

Users are allowed to override these beam parameters in the main macro by commenting out or adding more parameter calls following interface defined in [`PHHepMCGenHelper`](https://sphenix-collaboration.github.io/doxygen/dd/d2a/classPHHepMCGenHelper.html)

Given it changes behavior of of all event based simulation, this pull request is put in draft state. Help in cross checking the beam parameter calculation will be appreciated.

### DIFF
--- a/common/G4_Input.C
+++ b/common/G4_Input.C
@@ -94,7 +94,7 @@ namespace Input
     if (HepMCGen == nullptr)
     {
       std::cout << "ApplysPHENIXBeamParameter(): Fatal Error - null input pointer HepMCGen" << std::endl;
-      exit (1);
+      exit(1);
     }
     HepMCGen->set_beam_direction_theta_phi(1e-3, 0, M_PI - 1e-3, 0);  //2mrad x-ing of sPHENIX per sPH-TRG-2020-001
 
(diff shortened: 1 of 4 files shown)


Shuonli:
For the sPHENIX default macro, by getting the PHG4 truth particle information and plotting the spectator the beam distribution at z = 18m and z = -18m(where ZDC at), a beam position shift about 1.8 cm(1mrad) along the x direction is observed at both z position.
The same method is applied to the EIC default macro and a 25mrad shift in the positive z-direction is observed.
Outputs of both macros agree with our expectations.

PR #389, opened 2021-03-13
Updated FST and forward GEM configurations.

bschmookler:
This PR updates the current default EICDetector with an FST compatible with the TPC+MVTX+3 forward gems. The inner 4 FST disks are made smaller to fit inside the TPC inner field cage. The outer-most disk is moved to z=208cm to avoid a conflict with the last forward GEM. The inner radius of the second is decreased since no FST disk is nearby.

fst_fgem

In addition, an option for a 5 FGEM setting is created with an Enable::FGEM_ORIG option, which can only be enabled if the FST is removed.

original_fgem_full

Lastly, track projection to the EEMC is also added.

### DIFF
--- a/common/G4_FST_EIC.C
+++ b/common/G4_FST_EIC.C
@@ -30,6 +30,7 @@ namespace G4FST
     bool FSTV41 = false;
     bool FSTV42 = false;
     bool FSTV4 = false;
+    bool FST_MVTX_TPC = false;
   }  // namespace SETTING
 }  // namespace G4FST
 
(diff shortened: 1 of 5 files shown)


pingwong:
Hi Barak and Jin,

This is Ping at LANL, I am the major contributor to FST and central barrel macros. I understand that you may have an interest in another detector subsystem other than the FST and the current FST may not fit in your interested subsystem setup. However, it will be a mess for me and the LANL EIC team who have been working on the FST for almost a year if other groups change FST without consulting the LANL team. I hope you can understand my point of view. And I would like to ask you to push your version of FST, barrel and other related macros to your private branch (bschmookler:ephenix_update) without updating the FST and barrel macros in the master branch.

blackcathj:
Thanks for @bschmookler for completing the overlap checks. Merging to complete the forward tracking acceptance. 

Also thanks for @pingwong for commenting on resolving the conflictions with TPC. Looking forward to the improved FST setup  from  @pingwong soon too!

PR #391, opened 2021-03-16
latest FST and barrel designs (Rebased)

pingwong:
This pull request adds two FST designs and one barrel design. The latest version, version 5 of FST, consists of 6 silicon plane detectors with the last plane sits at z=300cm. Another FST design as shown in the following figure has 4 silicon planes located inside the TPC and 1 silicon plane outside the TPC. The 4 silicon plane has an outer radius of 17 cm, such that there is a 3 cm gap between the FST and the TPC for the FST service part.

The latest version (version 6) of the barrel tracker has 4 layers of silicon sensors. The outermost layer of the barrel tracks has a radius of 16cm, such that the barrel track can fit in the TPC.

To switch to the TPC-compactible FST and Barrel set

- G4FST::SETTING::FST_TPC=true
- G4BARREL::SETTING::BARRELV6=true

Screen Shot 2021-03-15 at 2 21 38 PM
Event display with the TPC (teal)+TPC endcap, FST (yellow) and the barrel (pink)


pingwong:
Hi @pinkenburg, I just rebased my branch. Let me if I did something wrong.

pinkenburg:
I re-enabled the barrel detectors again. For this macro we want all of the detector enabled. Users can then disable parts as they see fit. The DSTReader output is extremely large, it should be disabled by default so nobody accidentally overflows their disks when running over a few thousand events

PR #394, opened 2021-03-23
Macro for hadron far forward beamline and detectors

cipriangal:
This pull request implements the IP6 magnets and detectors following the design as created in eicROOT:
* magnets are read from the calibration folder
* 3 detectors: 1 ZDC plane, 2 Roman Pot planes, 4 B0 dipole planes

It also has a commented out beam pipe which has quite a few overlaps but may be useful to have if someone wants to fix it in the future. 

Figure of setup:
image

TODO:
* adjust/integrate with the large f4a framework (line 174 sets the world Z axis -- does this break anything?)
* generalize code to allow for IP6 or IP8 selection when calling the macro


pinkenburg:
It is a good start. needs to be modified so it can be called from a setup macro but this provides the needed ingredients so we can work on it

PR #397, opened 2021-04-02
integration with the main setup macro

cipriangal:
updated b0 detectors to not have a hole and brought the macros into the main setup. Both ip6 and ip8 configuration can be selected from flags. This will crash without a handling of the blackhole magnets in coresoftware (pull request pending)


detectors/EICDetector/Fun4All_G4_EICDetector.C
```diff
@@ -100,12 +101,12 @@ int Fun4All_G4_EICDetector(
...
 
   // eic-smear output
-  //  Input::READEIC = true;
+  Input::READEIC = true;
   INPUTREADEIC::filename = inputFile;
 
   // HepMC2 files
   //  Input::HEPMC = true;
-  Input::VERBOSITY = 0;
+  Input::VERBOSITY = 10;
```
blackcathj, line 109: Thanks for the update! Could you step back the commit that changes the default behavior of the macro: `const string &inputFile`, `Input::READEIC`, `Input::SIMPLE`, `Input::VERBOSITY`, etc. This help preserve behavior of users and avoids merge conflicts.

cipriangal:
Hey Jin (@blackcathj). This should be all set now. Thanks!

PR #412, opened 2021-05-12
remove other headers

osbornjd:
Remove other headers that is needed for the cleanup PR. Apparently my local tests are getting confused by the build libraries (which still contain these header files) and my local compiled libraries (which don't).


osbornjd:
@pinkenburg Jenkins in this [PR](https://github.com/sPHENIX-Collaboration/coresoftware/pull/1164) needs this macro change too. Apparently my local tests are getting confused by headers already installed in the build and installed in my local area.

PR #426, opened 2021-07-12
Added ZDC Digitization

Shuonli:
Construct ZDC towers


pinkenburg:
I had a brief look - the zdc tower map should be checked into the calibration repo (and then loaded from $CALIBRATIONROOT). The outer radius which is used for the black hole dimensions is set to 180. This is in cm - 1.8 m radius sounds large for the zdc, should this be 18 cm instead?

PR #427, opened 2021-07-14
added rudimentary direct and diffuse laser controls to g4_tpc

rccorliss:
Controls for Direct Laser and CM flash as implemented in the recent pull request to coresoftware.  All settings are within an if(false) so that there is (should be) no change in TPC behavior unless specifically changed.

TODO:
lasers should eventually live outside of the electron drift stage, and proper TPC ENABLE flags should replace the current hard-coded 'false'.


blackcathj:
superseded by #429

PR #440, opened 2021-09-21
Corrected the cone geometry

Shuonli:


pinkenburg:
This fixes from this PR were applied to PR 441 and the overlaps are gone, there is nothing wrong with this PR but merging this one would result in merging conflicts. Closing this PR

PR #451, opened 2021-10-01
Remove Acts vertexing

osbornjd:
We don't use the vertexing anymore and this module is slow in the `trackreco` compilation, so remove it here so that we can remove it from the `trackreco` makefile.


osbornjd:
Merged by Chris' [PR](https://github.com/sPHENIX-Collaboration/macros/pull/452), so closing this one.

PR #467, opened 2021-11-18
QA tracking mbhijing

osbornjd:
Add timers


pinkenburg:
This branch has significantly deviated by now from the master. It should be merged with the master before we proceed.

PR #468, opened 2021-11-18
include what you use

osbornjd:
Remove unused module


osbornjd:
@pinkenburg can we merge this so that I can push an empty commit for Jenkins to test this [PR](https://github.com/sPHENIX-Collaboration/coresoftware/pull/1343). This PR is just removing an include file from the G4_Tracking macro that is not used (and hasn't been for some time).

PR #469, opened 2021-11-23
Added HF QA and fast sim tracking service

cdean-github:
This PR enables HF QA for D0, Lambda_c and Upsilon decays as well as decay finder triggers for them all. The corresponding HF reco must run first so flags are in place which should ensure this can't crash things.

I'm still working on the full sim of the tracking service (almost there) but I had staged a fast sim without realizing which got pulled into this. I decided to do a little work on this so we can use the fast sim for the MDC2


cdean-github:
Screen Shot 2021-11-23 at 11 35 27 AM

PR #478, opened 2021-12-14
MVTX service update

cdean-github:
This PR adds the CYSS to the MVTX service material and connects the layers to the service barrel. Some small changes else where to avoid overlaps


cdean-github:
Screen Shot 2021-12-14 at 1 28 05 PM
Screen Shot 2021-12-14 at 11 23 33 AM

PR #498, opened 2022-03-15
Remove RAVE Dependence

osbornjd:
This PR removes all rave related code from the tracking macro as we do not use it anymore.


osbornjd:
The QA looks fine and thus I think we can merge this PR - however, this Jenkins build shows a good example of what we saw before. Removing RAVE from the tracking macro should not change the QA at all (certainly not the TPOT QA...) yet the histograms from the reference and new builds are not the same. The QA doesn't look egregiously different, but it is unclear why the histograms change at all @blackcathj

PR #515, opened 2022-05-03
G4 tracking split

hupereir:
This PR splits the G4_Tracking.C macro into a 
- seeding part: Tracking_Reco_TrackSeed()
- tracking part: Tracking_Reco_TrackFit()
This is necessary for writting Job A, B and C macros for our workflow. 
The possibility to run truth_track_seeding is kept. 
The original method "Tracking_Reco()" is kept to minimize disruption on existing macros. It is just a wrapper around the previous two methods

### DIFF
--- a/common/G4_Tracking.C
+++ b/common/G4_Tracking.C
@@ -335,10 +335,14 @@ void Tracking_Reco_TrackFit()
      * propagate tracks to EMCAL
      */ 
     
-    // Choose the best silicon matched track for each TPC track seed
-    auto cleaner = new PHTrackCleaner;
-    cleaner->Verbosity(verbosity);
-    se->registerSubsystem(cleaner);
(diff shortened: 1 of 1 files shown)


hupereir:
@blackcathj 
It might actually be useful to also merge this one to branches QA-tracking-low-occupancy and the like, so that it is run by Jenkins, and to double check that indeed I did not break anything. Does that make sense ? 
(I have tested the change in both normal and truth track seeding mode, but only for single particle simulations ...)

blackcathj:
Good point. Unfortunately, the `macros` appears already broken. Busy day today :)
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/sPHENIX_CoreSoftware_MasterBranch/1718/

PR #527, opened 2022-06-23
Global drift velocity

adfrawley:
Accompanies coresoftware PR#1547. This PR sets the global TPC drift velocity value in the G4_TPC.C macro and propagates it to all relevant tracking modules. The drift velocity is left at the presently used value of 8 cm / microsecond for now.


blackcathj:
The new macros work well with QA results reproduced at pValue=1. 
https://github.com/sPHENIX-Collaboration/macros/pull/527/checks?check_run_id=7099237755

Merging

PR #533, opened 2022-07-08
Revert "The phi offset of the INTT barrel changed"

pinkenburg:
Reverts sPHENIX-Collaboration/macros#532
This crashes our evaluator - it needs further investigation before we can use that

### DIFF
--- a/common/G4_Intt.C
+++ b/common/G4_Intt.C
@@ -49,7 +49,7 @@ namespace G4INTT
   int nladder[4] = {12, 12, 16, 16};
   double sensor_radius[4] = {7.188 - 36e-4, 7.732 - 36e-4, 9.680 - 36e-4, 10.262 - 36e-4};
 
-  double offsetphi[4] = {-0.5 * 360.0 / nladder[0], 0.0, -0.5 * 360.0 / nladder[2], 0.0 };
+  double offsetphi[4] = {0.0, 0.5 * 360.0 / nladder[1], 0.0, 0.5 * 360.0 / nladder[3]};
 
   enum enu_InttDeadMapType  // Dead map options for INTT
   {


nukazuka:
Did my change crush something? What can I do?

osbornjd:
From what I could tell, changing the offset in phi at the macro level somehow changes the hitsetkey generation in coresoftware. We have a mapping between Acts::Surface and hitsetkey that we use to lookup various things, and this change somehow broke that. So in addition to changing the offset in phi at the macro level, something (at this moment I'm not sure what) needs to be changed in coresoftware to ensure that the hitsetkey values are generated consistently. That was my best guess when I saw this crash

nukazuka:
Thank you for your explanation, Joe!

In my understanding, the default values of those phi offsets in coresofware repo are just 0:
https://github.com/sPHENIX-Collaboration/coresoftware/blob/8abe16b2b3d8567f1bea27b0643503f13662527e/simulation/g4simulation/g4intt/PHG4InttSubsystem.cc#L202

It's easy to put the offsets in the source code. Because I succeeded to the INTT model codes from the predecessor, I wonder why the offsets were not given in the source code although other parameters were given. I think it's time to fix the phi offsets. I try it if changing the source code can be a solution to the problem.

nukazuka:
Hi, I changed a code in coresoftware to set the phi offsets at that level, and it worked well without setting the offsets at the macro level. The changed code is simulation/g4simulation/g4intt/PHG4InttSubsystem.cc. The change is quite simple:

```
 $ git diff
diff --git a/simulation/g4simulation/g4intt/PHG4InttSubsystem.cc b/simulation/g4simulation/g4intt/PHG4InttSubsystem.cc
index 8f6df6dd4..2a30da8c4 100644
--- a/simulation/g4simulation/g4intt/PHG4InttSubsystem.cc
+++ b/simulation/g4simulation/g4intt/PHG4InttSubsystem.cc
@@ -189,6 +189,8 @@ void PHG4InttSubsystem::SetDefaultParameters()
                              7.188 - 36e-4, 7.732 - 36e-4, 9.680 - 36e-4, 10.262 - 36e-4,
                              12.676, 13.179};  // radius of center of sensor for layer default, new 30/05/2020

+  double offsetphi[4] = {-0.5 * 360.0 / nladder[0+2], 0.0, -0.5 * 360.0 / nladder[2+2], 0.0 }; // the final configuration, July/09/2022
+
   auto detid = GetDetIds();  // get pair of iterators to begin/end of set<int> of detids
   for (auto detiter = detid.first; detiter != detid.second; ++detiter)
   {
@@ -198,8 +200,10 @@ void PHG4InttSubsystem::SetDefaultParameters()
     set_default_int_param(*detiter, "laddertype", laddertype[*detiter]);
     set_default_int_param(*detiter, "nladder", nladder[*detiter]);  // ladders per layer
     set_default_double_param(*detiter, "sensor_radius", sensor_radius[*detiter]);
+
     // These offsets should be kept at zero in the new design
-    set_default_double_param(*detiter, "offsetphi", 0.);
+    //set_default_double_param(*detiter, "offsetphi", 0.); // obsolete
+    set_default_double_param(*detiter, "offsetphi", offsetphi[*detiter] );
     set_default_double_param(*detiter, "offsetrot", 0.);
```

If it's worth testing, I can make a pull request in the coresoftware repo.

PR #554, opened 2022-09-07
enable tpot

osbornjd:
The TPOT is not enabled by default. Is there a reason for this @pinkenburg ? This causes a crash in the tracking over the hits files when looking for the TPOT geometry.

### DIFF
--- a/common/GlobalVariables.C
+++ b/common/GlobalVariables.C
@@ -79,7 +79,7 @@ namespace XPLOAD
 
 namespace Enable
 {
-  bool MICROMEGAS = false;
+  bool MICROMEGAS = true;
 }
 
 namespace G4MICROMEGAS


pinkenburg:
Everything is disabled by default. The idea is that you enable what is needed in your macro to avoid things you don't want tp creep in automatically. The proper fix is to set this
Enable::MICROMEGAS = true;
in the macro which is crashing (you find the same in our MDC2 pass4_job0 to pass4_jobC macros)

hupereir:
> Okay, I understand your reasoning but this leads to some very cryptic seg faults in the tracking. Perhaps the better remedy is to put a warning in the geometry building in advance of these types of seg faults.

@osbornjd 
I agree, we should have some explicit messages about this in MakeActsGeometry. 
In fact this is also true if one tries to run the tracking with TPC, MVTX or INTT disabled, as far as I remember.

PR #566, opened 2022-11-03
remove sartre - it is retired

pinkenburg:
This PR removes the retired sartre from G4_Input.C


adfrawley:
Hi Chris,

This merge seems to have done before the new libraries were built, so Fun4All jobs fail because the Sartre include file is gone but the library code wants it.

Can you kick off a new build, and revert this for now until it completes?

Thanks
Tony

pinkenburg:
I removed sartre from the build yesterday but forgot about this macro which pulls it in. I just copied the new macro to the current offline build. The cvmfs release is going through, It should work again without changes on your side

PR #568, opened 2022-11-08
Removed default to truth seeding in pp

petermic:
Now that the CA seeder works in pp collisions, this removes the switch that defaulted to truth seeding in pp collisions.


petermic:
I removed the switch entirely, but it could simply be commented out -- let me know which is preferred.

osbornjd:
@blackcathj this does not need a CI check - we are removing the truth seeding from the streaming workflow because the TPC seeder now works for highly off vertex tracks.

PR #570, opened 2022-11-09
Commissioning function

osbornjd:
This PR adds a commissioning seeding function that can be called from the main macro level. The various parameters that could be further tuned are listed with initial values based on some short studies. The function will allow for further testing and tuning from a broader user base.


osbornjd:
Looks to me like everything passed, thanks @blackcathj for the new easy to use macros CI pipeline

PR #574, opened 2022-12-16
Introduce pp_mode switch for PHActsTrkFitter

petermic:
Introduces pp_mode for PHActsTrkFitter, which is necessary for TPC-only track fitting.


blackcathj:
A joint check of both PRs is started here: https://github.com/sPHENIX-Collaboration/macros/pull/574/checks?check_run_id=10144561427

PR #580, opened 2023-01-20
YCM using generic filename for the stave model symbolic link pointed …

ycorrales:
…to update geometry

This PR update the name of stave model file in calibration. Jenkins's test will fail until update of calibration repository in Jenkins.


ycorrales:
This will break simulations until update calibration repository. Wait for merge it

PR #589, opened 2023-02-13
Unifying beam parameter configuration selections for pp/pA/AA collisions

blackcathj:
Follow up to last software meeting discussion: unifying beam parameter configuration selections for pp/pA/AA collisions: 

* Explicitly select the nominal pp/pA/AA collision beam parameter configurations with `Input::BEAM_CONFIGURATION` in the main macro
* The same parameter is applied to event generator, hepmc inputs and pile ups, while it is possible to override for special studies
* Should be compatible with older user macros which default to emulate the nominal AuAu collision beam parameter configuration 
* The QA macro branches may need a sync update after merge to resolve merge conflicts

### DIFF
--- a/common/G4_Input.C
+++ b/common/G4_Input.C
@@ -81,27 +81,71 @@ namespace Input
   int VERBOSITY = 0;
   int EmbedId = 1;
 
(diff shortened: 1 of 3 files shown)


pinkenburg (approved):
looks good

PR #593, opened 2023-02-24
Add extra flag for epd tile

osbornjd:
This way if the EPD is turned off everything is turned off similarly to the other subsystems.


pinkenburg:
my bad - forgot that Thanks

PR #597, opened 2023-03-14
make cluster v5 default

bogui56:


osbornjd:
@bogui56 the QA looks reasonable and (as you suspected) the momentum resolution is improved. I suggest you or anyone else take a look before we merge this, but otherwise seems good to go:

https://github.com/sPHENIX-Collaboration/macros/pull/597/checks

PR #599, opened 2023-03-16
Remove MM function

osbornjd:
Forgot to add the commit which removes the function call to `MakeActsGeometry` regarding TPOT building.


osbornjd:
Build was with coresoftware https://github.com/sPHENIX-Collaboration/coresoftware/pull/1881, merging both

PR #606, opened 2023-03-24
2nd try of macro refactor

bogui56:
2nd try of macro refactor

Fun4All_G4 macros executes
Standalone clustering and tracking work and run at low memory usage.


osbornjd (approved):
@pinkenburg it looks like the CI passed. I would suggest we merge this and start a 2 week clock by which everyone has to switch their macros to the new scheme, after which we don't pledge support to the current G4_Tracking macro.

PR #610, opened 2023-04-02
Update to macros to work with PR1902 in coresoftware

david-stewart:
PR1902 in core software pulls in functionality also in PHG4InttDigitizer to PHG4InttTruthClusteriser (to use locally for the clusters for Truth Tracks). It therefore needs the G4_Intt.C macro to set the adc scales. This PR is needed to successfully run the code in PR1902 in coresoftware.

### DIFF
--- a/common/G4_Intt.C
+++ b/common/G4_Intt.C
@@ -11,6 +11,7 @@
 #include <g4intt/PHG4InttDigitizer.h>
 #include <g4intt/PHG4InttHitReco.h>
 #include <g4intt/PHG4InttSubsystem.h>
+#include <g4intt/PHG4InttTruthClusterizer.h>
 
 #include <g4main/PHG4Reco.h>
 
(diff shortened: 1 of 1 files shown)


osbornjd:
I kicked off a build using PR 1902, here's the link:
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/700/

david-stewart:
OK, I think it was an error in not including the required libraries in the G4_Intt.C macro, thereby causing the fails in the Valgrind, and presumably downstream effects. (I didn't get the error when running interactively).

It is now updated. At least interactively, it is running the default Fun4All_G4_sPHENIX.C macro without any complaints, so I think that it is good to go.

osbornjd:
Cool, build kicked off:
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/703/

PR #612, opened 2023-04-06
switching default ot new setup macros

bogui56:


osbornjd:
Testing with a build one more time after merging with master and inserting the relevant exit statements:

https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/810/

osbornjd:
I checked this manually with my local area and the QA failures are due to the ongoing issues with merging the new distortions QA into the default pipelines

PR #616, opened 2023-04-11
fix typo

hupereir:
Found this typo by browsing through the code.
@bogui56 @osbornjd 
This raises some question about the status of the macro rework. Are the "new" Trkr_XXX.C macros used already, in some steering macro  ? It would seem they are not, otherwise this code would not have compiled at all. Do I miss something ?


hupereir:
Note: G4_Tracking.C was written properly, with duplicated Distortions_QA method, but without the typo.

PR #618, opened 2023-04-12
changed pt range

hupereir:
title says it all


osbornjd:
I don't know if you are satisfied with a n=-4 power law, I just put that as a default to start such that the histograms would have a more realistic pT distribution. If you think it should be changed just add it to this PR, otherwise I'll merge it.

hupereir:
n=-4 seems to be good enough. It gives similar residuals to the one I get with an independent generator (measurements-based). So you can merge :)
In any case I am more interested in relative changes to the measured residuals from one commit to the other, than in the absolute value, as long as it is reasonable. 

One question though: is it worth turning off the TPC noise in this branch ? It speeds up the simulations significantly, and this noise does not bring much to the intended QA.

osbornjd:
> One question though: is it worth turning off the TPC noise in this branch ? It speeds up the simulations significantly, and this noise does not bring much to the intended QA.

Jin and I had talked about this - to be honest we should consider having this turned off by default anyway as it slows down development work. I think for now lets merge this and we can come back to this point as it affects all the QA branches. We'll have to update all of these anyway soon when we merge Christof's changes.

PR #621, opened 2023-04-14
TPC distortion update for new setup macros

bkimelman:
PR #612 readjusted macro setup and moved TPC distortions into Trkr_LaserClustering. This cannot be loaded at the same time as Trkr_Clustering, but Fun4All has a number of dependencies from Trkr_Clustering, so it breaks when trying to run with TPC distortions.Moved TPC distortion calls to Trkr_Clustering and removed Trkr_LaserClustering as no longer necessary which fixes this issue. The distortion calls require flags so they will not be used unless the flags are turned on.


osbornjd (commented):
Thanks for figuring out this fix and for submitting the PR for discussion. Can you explain why the `Trkr_LaserClustering` macro doesn't work, i.e. what it depends on that is in `Trkr_Clustering`? I'm having a hard time seeing exactly why this works (or the current  version doesn't) because your changes don't include the library for `tpccalib`. This might be an issue with the way our libraries are compiled or some other dependency, but I'm just trying to get a sense for what 'Trkr_LaserClustering' depends on that isn't there - is the Acts geometry, or the other clusterizers, or something else?

osbornjd:
Ah I just saw your mattermost comment. The problem is that the distortion related clustering still needs the silicon etc clustering. So it doesn't make sense to separate these. One could in principle just load all of the silicon libraries in `Trkr_LaerClustering` but that would then just be duplicated since `Trkr_Clustering` already does all of this. If my understanding is correct then I think that was an oversight when Christof was originally writing these macros and we should merge this. It only loads in one additional library (tpccalib)

bkimelman:
I don't think the distortion related clustering _actually_ needs the silicon and other clustering, but Fun4All_G4_sPHENIX.C does and it breaks whenever they aren't loaded, even if the flags are turned off. As you say, we could load those into Trkr_LaserClustering but then it would essentially do the same exact thing as Trkr_Clustering with the addition of the TPC distortions. I think it makes more sense to just add that functionality into Trkr_Clustering and remove Trkr_LaserClustering since it would essentially just be a duplicate

bkimelman:
Looking a bit more carefully, I think we do need some new flag since `Fun4All_G4_sPHENIX.C` does not have direct access to `G4_TrkrVariables`. Is there some other place to set the flags that could be accessed by `Fun4All_G4_sPHENIX.C` and then used to set the flags in `G4_TrkrVariables`?

bkimelman:
It looks like selecting which version of TPC_Clustering to run using the flags `G4TPC::ENABLE_DIRECT_LASER_HITS` and `G4TPC::ENABLE_CENTRAL_MEMBRANE_HITS` works. Should I change to this structure and update this PR?

osbornjd:
> That's what I did at first for my own personal "hack": I changed it to `#ifndef MACRO_TRKRLASERCLUS_C` and `TPC_LaserClustering`, which does work. I don't know if any new flag would be necessary, we can just use the ones for turning on the specific distortion functions to select which of the clustering macros to execute. I'm testing this now. If it works, should I change to this macros and update this PR?

Okay, I think if this works we should go this route. The goal of this whole reorganization was to keep things as separate as possible so that we can write individual fun4all macros that draw in only the bare minimum of what is needed.

> Looking a bit more carefully, I think we do need some new flag since `Fun4All_G4_sPHENIX.C` does not have direct access to `G4_TrkrVariables`. Is there some other place to set the flags that could be accessed by `Fun4All_G4_sPHENIX.C` and then used to set the flags in `G4_TrkrVariables`?

This is loaded in the default macro through `<Trkr_RecoInit.C>`, so all of the variables defined in TrkrVariables should be accessible through that

osbornjd:
> It looks like selecting which version of TPC_Clustering to run using the flags `G4TPC::ENABLE_DIRECT_LASER_HITS` and `G4TPC::ENABLE_CENTRAL_MEMBRANE_HITS` works. Should I change to this structure and update this PR?

Yes please, I think that will be the optimal solution to solve the problem as we can then maintain the separation of the macros but fix the current brokenness of the distortion workflow.

osbornjd:
Looks great, lets run Jenkins and see what happens. Thanks for the fix!

osbornjd:
Jenkins still is barfing with the npc issues from the last few days... we'll come back to this

PR #627, opened 2023-05-02
Refactor global vertex reco

osbornjd:
This PR accompanies
https://github.com/sPHENIX-Collaboration/coresoftware/pull/1990
which refactors the global vertex map object


osbornjd:
Here's a build
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/913/

PR #637, opened 2023-05-24
allow user to set the value of the event number

VVerkest:
This request is in conjunction with my request to edit g4eval/CaloEvaluator.h


pinkenburg (commented):
all good

PR #642, opened 2023-05-31
add plotting macro

eumaka:
PR adds a plotting macro to standardize approved plots plotting scheme. Initiating after recent plot approval process.


eumaka:
[plot1.pdf](https://github.com/sPHENIX-Collaboration/macros/files/11607603/plot1.pdf)

PR #645, opened 2023-05-31
also pass relevant TPC geometry parameters to PHG4TpcSubsystem

hupereir:
This is the first step to make the TPC geometry initialization work also when processing PRDFs.
We copy all the relevant parameters needed to define PHG4TpcCylinderGeom to PHG4TpcSubsystem. 
For now this should have no impact over running the simulations but is needed for a coming pull request to core_software.

### DIFF
--- a/common/G4_TPC.C
+++ b/common/G4_TPC.C
@@ -102,6 +102,12 @@ double TPC(PHG4Reco* g4Reco,
   tpc->SetActive();
   tpc->SuperDetector("TPC");
   tpc->set_double_param("steplimits", 1);  // 1cm steps
+
+  // copied from PHG4PadPlaneReadout
+  tpc->set_double_param("drift_velocity", G4TPC::tpc_drift_velocity_sim);
+  tpc->set_int_param("tpc_minlayer_inner", G4MVTX::n_maps_layer + G4INTT::n_intt_layer); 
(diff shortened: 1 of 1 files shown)


osbornjd:
@hupereir 
Please make the changes here, which is the simulation macro in the new scheme. We should probably remove these obsolete MVTX, INTT, TPC, TPOT macros as this is confusing right now.

https://github.com/sPHENIX-Collaboration/macros/blob/master/common/G4_TrkrSimulation.C

hupereir:
> @hupereir Please make the changes here, which is the simulation macro in the new scheme. We should probably remove these obsolete MVTX, INTT, TPC, TPOT macros as this is confusing right now.
> 
> https://github.com/sPHENIX-Collaboration/macros/blob/master/common/G4_TrkrSimulation.C
Will do. Thanks !

But now I am confused about how my checks worked in the first place. 
Possibly because the passed parameters are in fact identical to the default values ? 
I'll need to double check.

hupereir:
mmm but G4_Micromegas should not be loaded either. Right ? 
Cleanup seems to be in order yes :)

osbornjd:
> mmm but G4_Micromegas should not be loaded either. Right ? Cleanup seems to be in order yes :)

Ah right, duh. My mistake :D clearly cleanup is needed

hupereir:
I'll do a couple more tests, close this pull request, do a new one.

PR #646, opened 2023-05-31
Set the same parameters to PHG4TpcSubsystem as to PHG4TpcPadPlaneRead…

hupereir:
Set the same parameters to PHG4TpcSubsystem as to PHG4TpcPadPlaneReadout.
This is a first step towards moving the TPC initialization to PHG4TpcDetector.

(take two)

### DIFF
--- a/common/G4_TrkrSimulation.C
+++ b/common/G4_TrkrSimulation.C
@@ -272,6 +272,7 @@ void Intt_Cells()
 
 void TPCInit()
 {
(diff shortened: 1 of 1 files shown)


osbornjd:
@hupereir I assume this also needs to be merged? Should 2058 have also been tested with these changes as well?

hupereir:
This needs to be merged yes. However, there is no need to further test (I hope !): the change replaces default values by the values set in the macro, which are in fact the same ...

osbornjd:
Ah even better!

PR #663, opened 2023-08-09
Option in driving macro to do track matching w/ clusters

david-stewart:
May optionally generate a file with a TTree that recorded which track are matched (and which are not). May additionally put the clusteres of these tracks on the tree, along with which are, and are not, matched.

### DIFF
--- a/detectors/sPHENIX/Fun4All_G4_sPHENIX.C
+++ b/detectors/sPHENIX/Fun4All_G4_sPHENIX.C
@@ -58,11 +58,6 @@ int Fun4All_G4_sPHENIX(
   Fun4AllServer *se = Fun4AllServer::instance();
   se->Verbosity(0);
 
-  // Can enable writing out track matching (and track matching tree) here
-  // Enable::TRACK_MATCHING = true;
-  // Enable::TRACK_MATCHING_TREE = true;
-  // Enable::TRACK_MATCHING_TREE_CLUSTERS = true;
-
   //Opt to print all random seed used for debugging reproducibility. Comment out to reduce stdout prints.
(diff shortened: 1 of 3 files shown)


osbornjd (approved):
Here's a build which we can test just to be sure nothing changes, although it doesn't look like it will
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/1366/

detectors/sPHENIX/Fun4All_G4_sPHENIX.C
```diff
@@ -321,6 +326,11 @@ int Fun4All_G4_sPHENIX(
   Enable::TRACKING_EVAL = Enable::TRACKING_TRACK && true;
   Enable::TRACKING_QA = Enable::TRACKING_TRACK && Enable::QA && true;
 
+  // only do track matching if TRACKINGTRACK is also used
```
pinkenburg, line 329: This should be
Enable::TRACK_MATCHING = Enable::TRACKING_TRACK && false;
Enable::TRACK_MATCHING_TREE = Enable::TRACK_MATCHING && false;
Enable::TRACK_MATCHING_TREE_CLUSTERS = Enable::TRACK_MATCHING_TREE && false;
and remove the commented out lines 61-65. Then the user can activate this here (right after setting the tracking flags)

david-stewart: That is very clever! The changes are made in this next push.

osbornjd:
Thanks Chris, let's merge this

PR #665, opened 2023-08-11
editing Spacal to not call old functions that don't exist in new vers…

bseidlit:
…ion of PawClusterPositionCorrection.cc


pinkenburg:
The macro has a merging conflict. Rather than sorting this out here - closing it so it can be redone once this is sorted out

PR #673, opened 2023-09-18
Dat run23 qm

jtaou:


pinkenburg:
Most (all?) of the files are actual source code which should be compiled. This should be checked into our coresoftware repo where we have a calibrations/calorimeter area which already contains two modules (slope and tower by tower pi0). This should be moved there. I am closing this PR

PR #674, opened 2023-09-18
Dat run23 qm

jtaou:
all macros and code mods used for the summer 23/QM23 EMCal calibrations described in the corresponding analysis note.


jtaou:
sorry I saw Chris previous comment now

PR #685, opened 2023-10-18
Updating TPC Unpacking macro to include TpcCombinedRawDataUnpacker

rosstom2232:
Unpacks the prdf files using SingleTpcInput, then unpacks the TpcRawHits into the TrkrHitSetContainer using the new TpcCombinedRawDataUnpacker module


blackcathj:
Thanks, Thomas!

@pinkenburg has suggested us to split the unpacking part into a separate macro, e.g. `Fun4All_Tpc_Unpacking.C`, as we intend to run the combining step in a separate cycle in production. Could you split unpacking out and use the output from `[Fun4All_Tpc_Combiner.C](https://github.com/sPHENIX-Collaboration/macros/pull/685/files#diff-346589baafb56d2d89feb553d714aecc175873fa3c7381615b8c4cafd2af8b5b)` as its input?

PR #694, opened 2023-10-27
Changed INTT phi offset in G4_Trkr_Variables.C so it is now correct

josephbertaux:
This is the ladder phi offset fix Tony and I found a while ago; it should be trivial but give it a sanity glance over..

Nominally, this should be merged just before we re-run our next batch of simulations

### DIFF
--- a/common/G4_TrkrVariables.C
+++ b/common/G4_TrkrVariables.C
@@ -77,7 +77,7 @@ namespace G4INTT
   int nladder[4] = {12, 12, 16, 16};
   double sensor_radius[4] = {7.188 - 36e-4, 7.732 - 36e-4, 9.680 - 36e-4, 10.262 - 36e-4};
 
-  double offsetphi[4] = {0.0, 0.5 * 360.0 / nladder[1], 0.0, 0.5 * 360.0 / nladder[3]};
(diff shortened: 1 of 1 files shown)


pinkenburg:
Why are we setting these parameters in the macro? The macro overrides the parameters set in
https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/simulation/g4simulation/g4intt/PHG4InttSubsystem.cc
in  void PHG4InttSubsystem::SetDefaultParameters()
where the phi offset is actually correct. The macro then reset those to the wrong values which then botched our simulation. But then some values in PHG4InttSubsystem.cc are wrong. This scheme is extremely error prone and has to change.

What should be done is to fix PHG4InttSubsystem.cc to reflect what we have and leave the macro empty (or like the hcals leave correctly set parameters commented out to show what can be done if someone wants to experiment.

pinkenburg:
Closing this PR - the correct geometry is now set in the code

sPHENIX-GitHub-l:
Hi Joseph,

It is no longer needed to set the phi offsets from the macro, as Chris pointed out. The correct default setting for that is specified in PHG4InttDetector.

But there are still a number of parameters that must be set in the macro to get the correct INTT geometry. The default setup in the INTT code has 8 layers, and the macro changes that to 4 layers. This really should be modified so that the as-built geometry is run without any macro input needed. I looked at it but decided that I should leave it up to the INTT group.

Best regards
Tony

PR #712, opened 2023-11-13
ycm Add macro for MVTX standalone analysis

ycorrales:


pinkenburg:
This macro uses the outdated EvtPoolInput manager. Have you looked at https://github.com/sPHENIX-Collaboration/macros/blob/master/MvtxProduction/Fun4All_Mvtx_Combiner.C which does the same but uses the new streaming input manager

ycorrales:
> This macro uses the outdated EvtPoolInput manager. Have you looked at https://github.com/sPHENIX-Collaboration/macros/blob/master/MvtxProduction/Fun4All_Mvtx_Combiner.C which does the same but uses the new streaming input manager

No, I saw the new streaming input use the gl1_bco for matching something we don't need in the standalone, but now I see there is a different approach with this value set to zero. I will test this approach and then we can use the new streaming manager for both. Thank you

pinkenburg:
Have another look - the GL1 is only used if it is read in. Otherwise it runs standalone. The Fun4AllEvtInputPoolManager is superseded by the streaming input manager. It's not working properly and I am actually about to remove it.

ycorrales:
I have tested the streaming input and it return same result as previous Fun4AllEvtInputPoolManager using SetBcoRange to 0.
However the log output report processing an event less but the output root file contains all events. E.g for Run 25332 we expect 607490 events, log output print processing 607489 (index start from 1) but the root DST have correct number of events. Fun4AllEvtInputPoolManager reported the correct processing events, thought. I think we can just close this PR too

pinkenburg:
Thanks - I need to look into the logging - why an event is missing from the printout. But it's good to know that all of them make it to the DST

PR #721, opened 2023-11-21
feat: update unpacker to create output hitset DST

osbornjd:
Updates the unpacker macro to only create an output hitset dst. Copies the current macro to a new file which can still be checked in the CI. To be discussed at the software meeting today

### DIFF
--- a/TrackingProduction/Fun4All_TrkrHitSet_Unpacker.C
+++ b/TrackingProduction/Fun4All_TrkrHitSet_Unpacker.C
@@ -77,6 +77,14 @@ void Fun4All_TrkrHitSet_Unpacker(
   se->registerSubsystem(tpotunpacker);
 
   Fun4AllOutputManager *out = new Fun4AllDstOutputManager("DSTOUT", outfilename);
+  out->StripNode("MVTXRAWHIT");
+  out->StripNode("INTTRAWHIT");
+  out->StripNode("MICROMEGASRAWHIT");
+  out->StripNode("TPCRAWHIT");
+  out->StripNode("GL1RAWHIT");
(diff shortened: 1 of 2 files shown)


TrackingProduction/Fun4All_TrkrHitSet_Unpacker.C
```diff
@@ -80,46 +76,8 @@ void Fun4All_TrkrHitSet_Unpacker(
...
-  ntp->do_clus_trk_eval(true);
-  ntp->do_vertex_eval(false);
-  ntp->set_trkclus_seed_container("SvtxTrackSeedContainer");
-  ntp->Verbosity(0);
-  ntp->do_info_eval(false);
-  se->registerSubsystem(ntp);
-
-  //Fun4AllOutputManager *out = new Fun4AllDstOutputManager("out", "/sphenix/tg/tg01/hf/jdosbo/tracking_development/onlineoffline/hitsets.root");
-  //se->registerOutputManager(out);
+  Fun4AllOutputManager *out = new Fun4AllDstOutputManager("DSTOUT", outfilename);
```
blackcathj, line 79: We probably need to strip out the RawHit DST nodes here so the output is compact

osbornjd: [Commit](https://github.com/sPHENIX-Collaboration/macros/pull/721/commits/19128021952810c4a948d17830713dc754d4047c) that strips all nodes except the hitset node

osbornjd:
@blackcathj feel free to merge once you update the CI to call the new macro name that includes the full track seeding

PR #740, opened 2024-01-04
feat: add setter for extended readout for tpot

osbornjd:
Currently the TPOT extended readout is set as a default value but not overridden at the macro level, unlike the other detectors. This appears to make TPOT hits lost in the streaming simulations. This sets the minimum value to be the default value + the actual extended readout time set at the macro level.


osbornjd:
@hupereir I cannot request you as a reviewer for some reason so will you review this and advise if necessary.

osbornjd:
To add to this, this does naturally create hits within the extended readout time window that is set but 800 is taken directly from the default value currently in `PHG4MicromegasHitReco`. I'm not sure if this value is up to date or needs to be updated etc...

PR #764, opened 2024-03-07
Distortion flags

hupereir:
This PR adds a flag to select the format of the distortion corrections (namely using either radian or cm for the phi distortions), which is the propagated to the relevant modules. 

The default value is "true" which is also what's in the modules, so behavior is unchanged.

### DIFF
--- a/common/G4_Tracking.C
+++ b/common/G4_Tracking.C
@@ -57,7 +57,7 @@ void TrackingInit()
 {
 
   std::cout << "G4_Tracking is now obsolete as of macros PR612. "
-	    << std::endl << "Please switch to the new scheme for tracking, which is demosntrated in the PR at the following link: www.github.com/sPHENIX-Collaboration/macros/pull/612" 
+	    << std::endl << "Please switch to the new scheme for tracking, which is demosntrated in the PR at the following link: www.github.com/sPHENIX-Collaboration/macros/pull/612"
 	    << std::endl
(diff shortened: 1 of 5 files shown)


hupereir:
Side note: a lot of whitespace fixes have sneak in the commit. 
You can ignore them when looking at the file different using: https://github.com/sPHENIX-Collaboration/macros/pull/764/files?w=1

osbornjd:
Assuming this is intentional, but G4_Tracking is no longer used in our default tracking workflow. Not sure if you also need to update `Trkr_Reco.C`

osbornjd:
> Side note: a lot of whitespace fixes have sneak in the commit. You can ignore them when looking at the file different using: https://github.com/sPHENIX-Collaboration/macros/pull/764/files?w=1

Also... this is a cool feature, thanks for pointing it out

hupereir:
> Assuming this is intentional, but G4_Tracking is no longer used in our default tracking workflow. Not sure if you also need to update `Trkr_Reco.C`
Joe: I sort of know about some of the macros being obsolete, but just fixed nonetheless, in doubt and to keep things consistent. (also: I had doubts because of the new G4_Tracking_Cosmics.C)
I double checked that there is no need to fix Trkr_Reco. The relevant change for track reconstruction is in Trkr_RecoInit.

osbornjd:
Sounds good, just wanted to double check

PR #765, opened 2024-03-07
Fix truth reconstruction chain

hupereir:
set magnetic field parameter to relevant modules, to prevent exception when reading

### DIFF
--- a/common/Trkr_TruthReco.C
+++ b/common/Trkr_TruthReco.C
@@ -49,8 +49,8 @@ void convert_seeds()
   int verbosity = std::max(Enable::VERBOSITY, Enable::TRACKING_VERBOSITY);
 
   TrackSeedTrackMapConverter *converter = new TrackSeedTrackMapConverter();
-  // Default set to full SvtxTrackSeeds. Can be set to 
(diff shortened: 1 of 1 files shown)


hupereir:
Side note: 
@jdosbo 
@adfrawley 
I also turned ON actsFit->set_use_clustermover(true), because I need it (right ?) to properly handle distortion corrections.
Is that the right thing to do ? And if yes, should I also commit the same change to Trkr_Reco.C ? 

Hugo

adfrawley:
That is the default, so the macro should not need it.

I am in the process of trying to update the macros to properly run pp pileup events with distortions.

Tony

hupereir:
> That is the default, so the macro should not need it. I am in the process of trying to update the macros to properly run pp pileup events with distortions. Tony

in offline/packages/trackreco/PHActsTrkFitter.h, line 226: bool m_use_clustermover = false;

false is not the default (hence changing the flag in the macro. 
Should I change the default in the header to true ?

sPHENIX-GitHub-l:
Hah! Yes, please do. That was a mistake - I must have changed it for testing and forgot to set it back. My macro had it set to true, so I did not notice.
The other option does not work yet.

Thanks
Tony

PR #790, opened 2024-04-01
Add switches for INTT ideal/survey geometry

hrjheng:


osbornjd:
Here's a build
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/2524/

PR #792, opened 2024-04-10
Updated Trkr_Clustering to reflect changes in the larger PR

josephbertaux:


osbornjd:
Let me know when to create a build with this and I can, it should only affect the jenkins check with the hit unpackers and not the usual simulation jobs

josephbertaux:
I closed and resubmitted—I thought I missed a step in linking it to the coresoft PR. Chris already kicked off the new build for 793, we'll see what happens.

Joseph

PR #794, opened 2024-04-11
Distortion correction flags

hupereir:
Following today's discussion at the distortion meeting workgroup: 
this updates distortion correction flags to allow loading static distortion corrections and average distortion corrections independantly. 
For now, load from files.
Note (@adfrawley , @jdosbo, @pinkenburg etc.): 
if you have modified steering macros that change the correction flags, they will break 
Explicitly one must replace 
- G4TPC::ENABLE_CORRECTIONS  by G4TPC::ENABLE_STATIC_CORRECTIONS 
- G4TPC::correction_filename  by G4TPC::static_correction_filename

One can now also set and load ENABLE_AVERAGE_CORRECTION and average_correction_filename

TODO: 
- MDC macros might need modifications (I can do that if pointed to the right place)
- allow for using CDB instead of hardcoded root file path.

### DIFF
--- a/common/G4_Tracking_Cosmics.C
+++ b/common/G4_Tracking_Cosmics.C
@@ -70,8 +70,8 @@ void TrackingInit()
     auto tpcLoadDistortionCorrection = new TpcLoadDistortionCorrection;
     tpcLoadDistortionCorrection->set_read_phi_as_radians( G4TPC::DISTORTIONS_USE_PHI_AS_RADIANS );
 
-    if( G4TPC::ENABLE_STATIC_CORRECTIONS ) tpcLoadDistortionCorrection->set_correction_filename( 0, G4TPC::static_correction_filename );
-    if( G4TPC::ENABLE_AVERAGE_CORRECTIONS ) tpcLoadDistortionCorrection->set_correction_filename( 1, G4TPC::average_correction_filename );
+    if( G4TPC::ENABLE_STATIC_CORRECTIONS ) tpcLoadDistortionCorrection->set_correction_filename( TpcLoadDistortionCorrection::DistortionType_Static, G4TPC::static_correction_filename );
+    if( G4TPC::ENABLE_AVERAGE_CORRECTIONS ) tpcLoadDistortionCorrection->set_correction_filename( TpcLoadDistortionCorrection::DistortionType_Average, G4TPC::average_correction_filename );
 
     se->registerSubsystem(tpcLoadDistortionCorrection);
   }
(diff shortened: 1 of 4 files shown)


hupereir:
How does pushing changes to the macro repository work ? 
This one should go in. I don't think there should be any problem (famous last words)

hupereir:
It is also needed for https://github.com/sPHENIX-Collaboration/coresoftware/pull/2616

osbornjd:
The build I setup should have run with both PRs. Looking at the logfile 
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-detector-pipeline/7392/artifact/macros/detectors/sPHENIX/Fun4All_G4_sPHENIX.log/*view*/
it looks like it recognizes the new function but that the arg list is wrong. Was there some other update needed?

osbornjd:
Okay there were two builds, one that started automatically with the default macros and one that I started with this macro change. The load distortion map code indicates that the two arg list function is the correct one to call [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/dc45254631bcfa7851754d35c28a37312a98fed3/offline/packages/tpc/TpcLoadDistortionCorrection.h#L31-L43) so I'm not sure why it failed. Let me try to set it up again

hupereir:
ah crap. No the function call is wrong. Somehow I did not push all necessary commits. 
Let me try again.

hupereir:
ok. That should work now. 
Sorry

osbornjd:
I think it is still failing, assuming I set up the build last night correctly
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-detector-pipeline/7399/artifact/macros/detectors/sPHENIX/Fun4All_G4_sPHENIX.log/*view*/

hupereir:
Interesting, so your test already include the change from https://github.com/sPHENIX-Collaboration/macros/pull/794
And it fails with passing an int instead of an enum. 
This must be a 'feature' of CLING because the enum is explicitly said to derive from int, so it 'should' work.
Anyway. I'll fix.

hupereir:
all right. This one should work ... 
Sorry for the mess.

hupereir:
ok, so how can one re-run a build with these macros and the code included in https://github.com/sPHENIX-Collaboration/coresoftware/pull/2616 ? (which is needed for this macro to work)

osbornjd:
It is almost finished. See here
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/2614/

hupereir:
> It is almost finished. See here https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/2614/

Great ! Thanks !

PR #797, opened 2024-04-15
Follow up #795, enable the new field map

blackcathj:
Thanks to @pinkenburg , the new field map [[discussed in this notebook](https://nbviewer.org/github/sPHENIX-Collaboration/analysis/blob/master/FieldMap/FieldMapConversion_202404Extension.ipynb)] is now deployed through the CDB and CVMFS. This PR enable it. 

Note to @osbornjd , the tracking field variable is not enabled here yet, which I assume will be done with the tracking field map update in a separate step. 

Making this PR to allow for checking through Jenkins CI first....

### DIFF
--- a/common/G4_Magnet.C
+++ b/common/G4_Magnet.C
@@ -32,13 +32,11 @@ void MagnetFieldInit()
   }
   if (G4MAGNET::magfield.empty())
   {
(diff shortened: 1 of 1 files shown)


osbornjd (approved):
QA looks good to me, the changes are due to the INTT survey geometry and the reference has not been reset so that we don't lose track of what the track INTT measurement association rate was before the geometry was merged.

blackcathj:
Calo QA also look consistent. The field bending direction is also consistent with the past 
image

PR #801, opened 2024-04-17
Macro changes corresponding to coresoftware PR 2608

josephbertaux:
Applying a handle change for loading calibrations to the InttCombinedRawDataDecoder

### DIFF
--- a/InttProduction/Fun4All_Intt_Combiner.C
+++ b/InttProduction/Fun4All_Intt_Combiner.C
@@ -105,9 +105,9 @@ void Fun4All_Intt_Combiner(int nEvents = 0,
     InttCombinedRawDataDecoder *myDecoder = new InttCombinedRawDataDecoder("myUnpacker");
     myDecoder->runInttStandalone(true);
     myDecoder->writeInttEventHeader(true);
-    if (applyHotChannel) myDecoder->LoadHotChannelMapRemote("INTT_HotMap");
-    if (applyADCConversion) myDecoder->SetCalibDAC();
-    if (applyBCOCut) myDecoder->SetCalibBCO();
+    if (applyHotChannel)    myDecoder->SetCalib("INTT_HotMap", InttCombinedRawDataDecoder::FROM_CDB);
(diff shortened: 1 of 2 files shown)


josephbertaux:
Any thoughts on this and coresoftware 2608?

pinkenburg:
the valgrind errors are just formatting changes. These particular errors are suppressed via the line number in the source, if the line number changes they show up again. Once the dust settles, the suppression file will be adjusted

josephbertaux:
The default loading behavior comes from InttCombinedRawDataDecoder. One of the issues with loading things by default is that test builds may fail if the corresponding tags are not found in the CDB (the loading method complains and exits with return value 1). I wanted to load our calibrations by default with the correct default tags, but had to undo it b/c there were no entries for the BCO or DAC map for the particular timestamp Jenkins used and this caused the test build to fail. The low-hanging fix is to make sure there are at least dummy versions of those calibrations for the test macro, but then we can't actually use it until those calibrations are in for the runs we want to analyze, which could be a production setback and "failing" in this way is desirable to flag as an issue.

I can change the handles from something like SetCalib(<tag>, <method>, <file>) to UseBcoMap(<file/tag>="INTT_BCOMAP"), etc. and call whichever helper function based on if the argument contains ".root" or skip that calibration if the name is passed as an empty string. There'd still be the issue of maintaining the default values, wherever they are specified, but at least that's not something users need to worry about.

It seems like, on the macros side, the calibrating of the unpacker should be done [here](https://github.com/sPHENIX-Collaboration/macros/blob/4afff83af4e59b7695717e9f572a11469e303486/common/Trkr_Clustering.C#L65-L74), and then I can make code that relies on the unpacker use it (e.g. [this block](https://github.com/sPHENIX-Collaboration/macros/blob/4afff83af4e59b7695717e9f572a11469e303486/InttProduction/Fun4All_Intt_Combiner.C#L105-L111)). As for specifying tag names, that can probably be done [here](https://github.com/sPHENIX-Collaboration/macros/blob/4afff83af4e59b7695717e9f572a11469e303486/common/G4_TrkrVariables.C#L83-L89). That section of code should be replaced, anway--it isn't (or at least, shouldn't) be doing anything.

PR #808, opened 2024-04-19
Reconciled INTT calibration handles

josephbertaux:
Collected configuration flags being used in the InttProduction macro and moved them to G4_TrkrVariables and Trkr_Clustering; updated the methods in Trkr_Clustering to reflect this, and use the updated method in the unpacking.

If  the removal of the enu_InttDeadMapType "breaks" anything, whatever depended on it should not be used--to my knowledge that was never correct or meaningful.

### DIFF
--- a/InttProduction/Fun4All_Intt_Combiner.C
+++ b/InttProduction/Fun4All_Intt_Combiner.C
@@ -59,12 +59,13 @@ void Fun4All_Intt_Combiner(int nEvents = 0,
 
   //! [TO CONFIRM] The switches for the ideal and survey geometries do not seem to work anymore. Is this intended?
   Enable::INTT_USEG4SURVEYGEOM = usesurveygeom;
+  ACTSGEOM::inttsurvey = usesurveygeom;
 
   TString outname = Form("ProdDST/intt-%08d", runnumber);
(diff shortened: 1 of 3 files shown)


josephbertaux:
I've marked this as a draft, since it's worth checking that key macros still work if this gets merged (namely, InttProduction/Fun4All_Intt_Combiner.C)

josephbertaux:
I did some poking around--omitting the dead channel loading enum affects G4_TrkrSimulation--it uses these to add an instance of PHG4InttDeadMapLoader to the nodeTree (which does nothing). I looked in g4intt, and I can't see where instances of the InttDeadMap are actually used by other classes (grep only shows results for InttDeadMap in the source files that define the class and the loader--I was at least expecting something else to try to get an instance from the node tree).

I can leave the enum in for now, but eventually the PHG4Intt things are going to need to be modified to use the same thing the unpacker is using, and that's more than what I want to do in this PR

josephbertaux:
This is now tied with coresoftware PR 2639; I'll let the automatically queued checks fail and then start the build from here

Also, the dead channel masking should be done in simulations--I'd like to do this, but in a separate PR. This one is already taking too much time.

PR #809, opened 2024-04-19
temp file bug fix

bseidlit:


pinkenburg:
We really don't want pointers to files in user directories, this defeats the purpose of this being runnable under the sphnxpro account. Consider making the file a function argument. Same for the relative path ../, if this becomes part of a condor job, this will likely not work

PR #827, opened 2024-05-02
Diffuse Laser Clusterizer

bkimelman:
Default to now using new laser clusterizer and matching algorithm


osbornjd:
Can you reopen this PR with only the files that have actual changes in them? The default macro has some spacing changes and  sometimes jenkins has a hard time with things like this if git identifies merge conflicts, so if we can avoid it from the start that will be easier

bkimelman:
> Can you reopen this PR with only the files that have actual changes in them? The default macro has some spacing changes and sometimes jenkins has a hard time with things like this if git identifies merge conflicts, so if we can avoid it from the start that will be easier

Done

osbornjd:
Thanks!

PR #851, opened 2024-05-17
Fix tpc geometry

hupereir:
This makes sure that wathever value is set to G4TPC::tpc_layer_rphi_count_inner is applied consistently to both PHG4TpcSubsystem and PHG4TpcPadPlaneReadout. 
@mohaas33 please double check the value.


hupereir:
@mohaas33 ping ?
I'd like to get this merged

mohaas33 (approved):

PR #866, opened 2024-06-14
Added residual corrections

mitrankova:
Enabled static correction flag with the path to the residual correction map


pinkenburg:
Is there a chance that we stop using private directories for the distortion corrections? We have space in cvmfs or - even better - add this to the cdb

PR #877, opened 2024-06-26
Calo Y2 arguments and further cosmetic changes

bseidlit:


SashaLebedev:
I'm not completely sure, but I would keep fulloutfile and fulloutfile_hist everywhere because they have "./" in front. It could make a difference in bash, as opposed to tcsh, but I'm not really an expert on this.

pinkenburg:
Why would you do this? If you pass down a fully qualified filename (with path - e.g. /sphenix/user/myname/hist.root), you prepend ./ resulting in .//sphenix/user/myname/hist.root which will not work. If you just give a filename it will be opened in the local directory (by every shell) with or without ./

CaloProduction/Fun4All_Year2.C
```diff
@@ -63,17 +68,10 @@ void Fun4All_Year2(const std::string &fname = "DST_TRIGGERED_RAW_beam_new_2023p0
...
-  string fulloutfile_hist = string("./") + outfile_hist;
-
-  // conditions DB flags
-  rc->set_StringFlag("CDB_GLOBALTAG", "ProdA_2024");
-  // 64 bit timestamp
+  //int segment = runseg.second;
+
+  // conditions DB flags and timestamp
+  rc->set_StringFlag("CDB_GLOBALTAG", dbtag);
   rc->set_uint64Flag("TIMESTAMP", runnumber);
```
pinkenburg, line 75: This should be removed. The TIMESTAMP is set by the framework.

PR #888, opened 2024-07-06
change order of call for zdc reco

eumaka:


pinkenburg:
Is there a reason to run the zdc after the global vertex? I thought the plan is to get a vertex from the zdc in which case the global vertexing needs to be registered after the zdc reco.

eumaka:
That's right Chris-- full zdc reco now before global vertex map

PR #899, opened 2024-07-19
production macros with tpc reco init

bogui56:


pinkenburg:
This PR needs more discussion. Passing the run number as an argument is really not what we want. This will break too many things downstream (also the production)

osbornjd:
I was under the impression that the production scripts ran the macros from [here](https://github.com/klendathu2k/slurp-examples/tree/main/sPHENIX)? All of the streaming/tracking macros here take the run number as an argument

bogui56:
I submitted the macro changes here so we can first of all run it through jenkins and then review/discuss how to properly move the changes to the production area.
Please have a look and advise...

pinkenburg:
Resolved merging conflict by hand and merged this PR

PR #902, opened 2024-07-23
Added some flags to run different TPC gas mixtures

cdean-github:
This PR should match with https://github.com/sPHENIX-Collaboration/coresoftware/pull/3002

This will set the TPC to use 75:20:5 Ar:CF4:isobutane as its default gas mixture. Some strings exist to use other mixtures


blackcathj:
Thanks! Here is the joint check on both PRs: https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/3430/

PR #906, opened 2024-08-02
yuxd: Add Fun4All macro for track+calo study

yuxdPKU:
We add a new Fun4All macro to do track and calo related analysis.
- tracking
- track projection
- topo cluster reconstruction
Also provide option to do
- TPC-only tracking or full tracking
- TPC drift velocity correction
- EMcal radius correction in track projection

### DIFF
--- a/TrackingProduction/Fun4All_FieldOnAllTrackersCalos.C
+++ b/TrackingProduction/Fun4All_FieldOnAllTrackersCalos.C
@@ -1,0 +1,411 @@
+/*
+ * This macro shows a minimum working example of running the tracking
+ * hit unpackers with some basic seeding algorithms to try to put together
+ * tracks, calorimeter cluster reconstruction, and track projected to claorimeter surface. There are some analysis modules run at the end which package
+ * hits, clusters, and clusters on tracks, calo clusters, and towers in calo clusters into trees for analysis.
+ * also provide the option to do full tracking detectors or TPC only, and drift velocity correction and EMcal radius correction
+ */
+
+#include <G4_ActsGeom.C>
+#include <G4_Magnet.C>
+#include <GlobalVariables.C>
+#include <Trkr_Clustering.C>
+#include <Trkr_RecoInit.C>
+#include <Trkr_Reco.C>
+#include <G4_Global.C>
+
+#include <caloreco/RawClusterBuilderTopo.h>
+
+#include <ffamodules/CDBInterface.h>
+#include <fun4all/Fun4AllDstInputManager.h>
+#include <fun4all/Fun4AllDstOutputManager.h>
+#include <fun4all/Fun4AllInputManager.h>
+#include <fun4all/Fun4AllUtils.h>
+#include <fun4all/Fun4AllOutputManager.h>
+#include <fun4all/Fun4AllRunNodeInputManager.h>
+#include <fun4all/Fun4AllServer.h>
+
+#include <phool/recoConsts.h>
+
+#include <track_to_calo/TrackCaloMatch.h>
+#include <track_to_calo/TrackToCalo.h>
+#include <track_to_calo/CaloOnly.h>
+#include <track_to_calo/TrackOnly.h>
+
+#include <trackreco/AzimuthalSeeder.h>
+#include <trackreco/PHActsTrackProjection.h>
+#include <trackingdiagnostics/TrackResiduals.h>
+
+#include <trackbase_historic/SvtxTrack.h>
+
+#include <stdio.h>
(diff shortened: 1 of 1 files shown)


TrackingProduction/Fun4All_FieldOnAllTrackersCalos.C
```diff
@@ -0,0 +1,411 @@
...
+  std::pair<int, int> runseg = Fun4AllUtils::GetRunSegment(firstfile);
+  int runnumber = runseg.first;
+  int segment = runseg.second;
+
+  auto rc = recoConsts::instance();
+  rc->set_IntFlag("RUNNUMBER", runnumber);
+
+  Enable::CDB = true;
+  rc->set_StringFlag("CDB_GLOBALTAG", "ProdA_2024");
+  rc->set_uint64Flag("TIMESTAMP", 6);
```
pinkenburg, line 87: This line will break any cdb lookup. It will use the calibrations for run number 6 (which is a simulation run). Any runwise calibration for the e.g. calos will be wrong

TrackingProduction/Fun4All_FieldOnAllTrackersCalos.C
```diff
@@ -0,0 +1,411 @@
...
+#include <fun4all/Fun4AllUtils.h>
+#include <fun4all/Fun4AllOutputManager.h>
+#include <fun4all/Fun4AllRunNodeInputManager.h>
+#include <fun4all/Fun4AllServer.h>
+
+#include <phool/recoConsts.h>
+
+#include <track_to_calo/TrackCaloMatch.h>
+#include <track_to_calo/TrackToCalo.h>
+#include <track_to_calo/CaloOnly.h>
```
osbornjd, line 32: Are these private libraries? We should remove them so that anyone can run this out of the box

TrackingProduction/Fun4All_FieldOnAllTrackersCalos.C
```diff
@@ -0,0 +1,411 @@
...
+  string outputDstFileName = "TrackCalo_" + to_string(segment) + "_dst.root";
+
+  string outputRecoDir = outDir + "/inReconstruction/" + to_string(runnumber) + "/";
+  string makeDirectory = "mkdir -p " + outputRecoDir;
+  system(makeDirectory.c_str());
+  string outputDstFile = outputRecoDir + outputDstFileName;
+
+  std::cout << "Output dst file: " << outputDstFile << std::endl;
+
+  Global_Reco();
```
osbornjd, line 118: This should go at the end of processing, after the tracking vertexing is done

TrackingProduction/Fun4All_FieldOnAllTrackersCalos.C
```diff
@@ -0,0 +1,411 @@
...
+    dz_separation_0 = -0.13*2*105; // recommened drift velocity correction for Ar/CF4/N2 gas
+  }
+  else
+  {
+    dz_separation_0 = 0;
+  }
+  G4TPC::tpc_drift_velocity_reco = (8.0 / 1000) * 107.0 / 105.0 * (1+ dz_separation_0 / 2. / 105.); //cm/ns
+  G4TPC::ENABLE_MODULE_EDGE_CORRECTIONS = true;
+  //to turn on the default static corrections, enable the two lines below
+  G4TPC::ENABLE_STATIC_CORRECTIONS = true;
```
osbornjd, line 138: Let's leave this to false right now. We don't understand the static corrections well enough at this point to have people running with them

osbornjd (changes requested):

eumaka:
It would be help to add a short list of the contents of : "run46730_0000_trkr.txt", "run46730_calo.list" as an example.

TrackingProduction/Fun4All_FieldOnAllTrackersCalos.C
yuxdPKU: Here, I refer to Fun4All_FieldOnAllTrackers.C. Do you have any suggestions to modify?

osbornjd: It needs to be updated there too. We should set it to the actual run number

yuxdPKU: `  Enable::CDB = true;
  rc->set_StringFlag("CDB_GLOBALTAG", "ProdA_2024");
  rc->set_uint64Flag("TIMESTAMP", runnumber);`
  Is it okay?

PR #911, opened 2024-08-14
TPC residuals

hupereir:
Added relevant modules to have residuals in the TPC measured with TPC clusters taken out of the ACTS fit.
To enable this one must

- change convertSeeds to false (in the macro arguments) in order to enable full track fit with ACTS
- G4Tracking::SC_CALIBMODE to true (in the macro body) in order  to remove TPC clusters from the ACTS fit and add the PHTpcResiduals module, responsible for extrapolating the acts fit track parameters to the TPC layers that contain a cluster.

### DIFF
--- a/TrackingProduction/Fun4All_FieldOnAllTrackers.C
+++ b/TrackingProduction/Fun4All_FieldOnAllTrackers.C
@@ -27,6 +27,8 @@
 
 #include <phool/recoConsts.h>
 
+#include <tpccalib/PHTpcResiduals.h>
+
 #include <trackingqa/InttClusterQA.h>
 #include <trackingqa/MicromegasClusterQA.h>
 #include <trackingqa/MvtxClusterQA.h>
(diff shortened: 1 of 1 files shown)


osbornjd:
Looking this over made me wonder - could we write two track residual modules and run both the full and silicon+TPOT fitting in the same job, since they write to different track maps, and have the track residual trees generated for both at the same time? Might need some changes in coresoftware to be able to handle but that might be useful to have both in the same output file...

hupereir:
@osbornjd: we could, and I don't think there is any change needed to core software. Just register two ACTS fitter, with opposite SC_CALIB_MODE flags (they then write to two different track nodes automatically), and two TpcResiduals with tweaked flags. 
Still: that  will complexify the macro quite significantly, for something that ultimately is temporary. 
Can we commit this first, unchanged, and discuss this at tomorrow's meeting ?

PR #915, opened 2024-08-21
move drift velocity calib macro from coresoftware to macro

yuxdPKU:


osbornjd:
These are all macros required to run things in the semi-automated way that we discussed?

yuxdPKU:
Yes, i think so

PR #920, opened 2024-08-23
Fixed Laser Clustering

bkimelman:
Fixed issue with laser clustering being in two different macros. Removed it from Trkr_Clustering.C and updated version in Trkr_LaserClustering.C for use


osbornjd:
Does anyone else use these other laser clustering modules (i.e. do we want to just remove them entirely)?

bkimelman:
> Does anyone else use these other laser clustering modules (i.e. do we want to just remove them entirely)?

I don't think so. I'll check with Charles

bkimelman:
> > Does anyone else use these other laser clustering modules (i.e. do we want to just remove them entirely)?
> 
> I don't think so. I'll check with Charles

Charles said it's fine to remove. It also makes sense as this is just for clustering. The analysis of the laser should be in a different macro that doesn't have to load the clustering libraries, which we can manage later

osbornjd:
This macro change needs https://github.com/sPHENIX-Collaboration/coresoftware/pull/3089 to go through first

PR #922, opened 2024-08-27
feat: add tpc zero supp option

osbornjd:
Adds an optional flag to set the zero suppression hit unpacking option in the tpc


osbornjd:
Merging this as it passes some of the basic macros, which means it doesn't break anything

PR #934, opened 2024-09-09
allowing radian vs cm control for both static and average maps

rccorliss:
We didn't anticipate using the simulated average spacecharge correction, which is intended to come from direct processing of data, so this pull req adds flags to control reading those files, which were not previously present in the macros.  This also harmonize the naming of the flags used to control the input type for the static correction with those used for the average correction.


osbornjd:
It appears that it is commented out in `G4_TrkrVariables.C` - is that intentional?

rccorliss:
> It appears that it is commented out in `G4_TrkrVariables.C` - is that intentional?

It was intentional, but in error.  I restored it, and added a note about what that flag is still used for.  Thank you for commenting.

PR #946, opened 2024-09-21
Change silicon-TPC matcher settings and add track cleaner in Fun4All_FieldOnAllTrackers.C

adfrawley:
Changes settings of PHSiliconTpcTrackMatching in Fun4All_FieldOnAllTrackers.C. Removes setting of the "use_old_matching" flag - it is only for testing, not for normal use. Changes the pp_mode flag setting so it is not hard coded to true. 

Adds PHTrackCleaner after the Acts track fitter.

### DIFF
--- a/TrackingProduction/Fun4All_FieldOnAllTrackers.C
+++ b/TrackingProduction/Fun4All_FieldOnAllTrackers.C
@@ -238,8 +238,7 @@ void Fun4All_FieldOnAllTrackers(
   silicon_match->set_z_search_window(5.);
   silicon_match->set_phi_search_window(0.2);
   silicon_match->set_eta_search_window(0.1);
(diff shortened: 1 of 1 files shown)


osbornjd (commented):
Removing `set_use_old_matching` will cause us to use windows that expand as a function of pT and cause you to get many many matches at low pT, where most of our tracks are. Did you test this to see what happens?

sPHENIX-GitHub-l:
Let's hold off on merging it for now.

I tested it on run 53199, and it increases the number of matched tracks (including above pT = 0.2) very significantly. But I am making other changes to the macro too, looking at how to maximize the matching yield. I am also modifying the reco chain to make the non pp mode behavior better.

I will update this PR and make a coresoftware PR when I am done.

PR #947, opened 2024-09-22
fix: extend strobe to 10 mus

osbornjd:
We didn't run any physics data with a 5 mus strobe, so setting 10 mus as the default for now.


cdean-github:
This looks good to me. Technically the strobe is 9.9 us but this should have no impact on our simulations

cdean-github (approved):

PR #956, opened 2024-10-01
MVTX modified geometry (macros)

hrjheng:
- Add boolean variables to use the default ideal or modified (misaligned) MVTX geometry 
- G4_Pipe.C: move the beampipe according to the survey so fix the overlap between MVTX and the beampipe

Corresponding PR for the software and calibrations:
- coresoftware:
- calibrations: https://github.com/sPHENIX-Collaboration/calibrations/pull/158


osbornjd:
There are still changes to the QA as indicated by the p values
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/3848/
But they aren't huge. Maybe that is good enough to go forward? Do you know if there was anything else that wasn't hidden behind the flag?

hrjheng:
I can only think of and find one place in PHG4MvtxDetector where the change is not behind the flag: https://github.com/sPHENIX-Collaboration/coresoftware/pull/3166/commits/6eeeb445ccd85fca3e419cdc9241bba9f7493c3f
I am not sure if this will change the QA, especially the ones for micromegas and tpc that this PR does not touch

osbornjd:
It might because the material would be different, so that could impact what happens in the g4 sim. I'll setup a new build with the latest pushes

hrjheng:
@osbornjd I have an update. After a message exchange with Dennis and a discussion during the tracking meeting today, the conclusion is for me to: 
1. Update the preliminary local alignment parameters by subtracting the global displacements used in the current G4 geometry.
2. Update the alignment parameter files in CDB (I'll send the file to Chris once 1 is done) and set all switches in macros to true to use misaligned MVTX geometry
This should guarantee that (a) the simulation geometry and the relevant part of the reconstruction are updated together and (2) we'll have a geometry and detector material placement closer to reality for the calo simulation production that is going to  start soon

PR #962, opened 2024-10-04
(calibrations/intt) Cleanup

josephbertaux:
Simplified run.sh, retrieve files through InttOdbcQuery instead of ls, removed condor scripts


pinkenburg:
The current macro is ready to go, the input files will be handed down from the production system - closing this PR

PR #978, opened 2024-10-21
djl -- added trigger runinfo to year2 macro

danjlis:
This adds the trigger run info to the Year2 macro. Before anything else to ensure any module that wants to use the trigger info has access to the TriggerRunInfo node. 

At this moment, it prints out the trigger menu with names, prescales and scalers to the output file.


bseidlit:
We have split the Year2 macro up into fitting and calib.  I think this would fit best in fitting since it will not change.  Can you also put it in the fitting macro.  Maybe we should get ride of this combine year2 macro to avoid confusion.

PR #981, opened 2024-10-23
Mvtx triggered mode flag

hrjheng:
Added a switch to set MVTX in the triggered mode for HitUnpacking
Corresponding software part: https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/packages/mvtx/MvtxCombinedRawDataDecoder.cc

### DIFF
--- a/common/Trkr_Clustering.C
+++ b/common/Trkr_Clustering.C
@@ -4,11 +4,11 @@
 #include <GlobalVariables.C>
 
 #include <G4_ActsGeom.C>
-#include "G4_TrkrVariables.C"
+#include <G4_TrkrVariables.C>
 
 #include <intt/InttCombinedRawDataDecoder.h>
 #include <micromegas/MicromegasCombinedDataDecoder.h>
-#include "mvtx/MvtxCombinedRawDataDecoder.h"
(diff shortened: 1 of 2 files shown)


common/Trkr_Clustering.C
```diff
@@ -4,11 +4,11 @@
 #include <GlobalVariables.C>
 
 #include <G4_ActsGeom.C>
-#include <G4_TrkrVariables.C>
+#include "G4_TrkrVariables.C"
```
osbornjd, line 7: These need to be left as <> to by default look for the nightly libraries

common/Trkr_Clustering.C
```diff
@@ -4,11 +4,11 @@
...
 
 #include <intt/InttCombinedRawDataDecoder.h>
 #include <micromegas/MicromegasCombinedDataDecoder.h>
-#include <mvtx/MvtxCombinedRawDataDecoder.h>
+#include "mvtx/MvtxCombinedRawDataDecoder.h"
```
osbornjd, line 11: Same here

osbornjd (changes requested):

osbornjd (approved):

PR #992, opened 2024-11-09
Extend Suite of Jet Workflows

ruse-traveler:
This PR extends the suite of possible jet reconstruction workflows to include non-background-subtracted calorimeter, track, and particle-flow jets, and to include background-subtracted track and particle-flow jets. This will help centralize jet reconstruction for non-background-subtracted  and track-based analyses, and step towards integrating track jets into the jet QA.

For the former point, a new macro, `NonBkgdSubJetReco.C`, has been added to the common macros. And for the latter, `HIJetReco.C` has been reorganized to allow the user to run jet reconstruction on tracks and particle-flow elements.

Note that since background-subtraction routines for track jets and particle-flow jets are still under development, the routines in `HIJetReco.C` are placeholders that should be filled in later. The macro will emit a warning saying as much.

**Note:** requires [coresoftware#3488](https://github.com/sPHENIX-Collaboration/coresoftware/pull/3488)


common/HIJetReco.C
```diff
@@ -29,6 +32,9 @@ namespace Enable
   int HIJETS_VERBOSITY = 0;
   bool HIJETS_MC = true;
   bool HIJETS_TRUTH = false;
+  bool HIJETS_TOWER = true;   // make tower jets
+  bool HIJETS_TRACK = false;  // make track jets
+  bool HIJETS_PFLOW = false;  // make particle flow jets
 }  // namespace Enable
```
ruse-traveler, line 38: TO-DO: add doBkgdSub flag?

ruse-traveler:
Summary of discussion w/ JS TG and offline discussion w/ @KvapilJ , @xuanli23 , @zhiwanxu:
- The `HIJetReco` macro is specifically for background subtracted jets, so it could be confusing to mix background-subtracted & non-background-subtracted workflows. One fix could be to add a `doBkgdSub` flag and emit a warning that background subtraction isn't implemented yet for charged/PF jets when running those.
- However, there are use cases (e.g. HF-tagged jets) where doing no background subtraction is preferable. It might be useful to have a macro explicitly for such workflows

common/HIJetReco.C
ruse-traveler: Since there use-cases for running with jet reconstruction without background subtraction, I opted to instead add a parallel "no subtraction" macro, `NoBkgdSubJetReco.C`. My hope is that this will make it more clear what workflow each of these macros is for.

For `HIJetReco`, though, I've added warnings that track/PF background subtraction is still under development if the user tries to run track/PF jet reconstruction in HIJetReco.

PR #999, opened 2024-11-16
Revert "Revert "CD: Added more beam pipe info""

pinkenburg:
This puts the modified beam pipe macro back so we can check it out. It crashes likely because a duplicate detector (volume) id. Marking this draft so it won't be merged accidentally


cdean-github:
I pushed a patch but forgot this was a revert so it didnt propagate. The offending line is here https://github.com/sPHENIX-Collaboration/macros/blob/master/common/G4_BeamLine.C#L45

This is hardcoded to 100. I switched it to 300 (as there's another hardcoding of 200 immediately below). Maybe we need a global variable to count these objects to avoid clashes in the future

pinkenburg:
should be fixed in pr 1006

PR #1003, opened 2024-11-22
Added Fun4All_TrackSeeding.C

aopatton:
Added a macro that runs Track Seeding over a cluster DST

### DIFF
--- a/TrackingProduction/Fun4All_TrackSeeding.C
+++ b/TrackingProduction/Fun4All_TrackSeeding.C
@@ -1,0 +1,432 @@
+/*
+ * This macro shows a working example of running TrackSeeding over the cluster DST
+ * This has track residuals as default output but has KFParticle set up with a togglable flag
+ * with the default set up for K Short reconstruction
+ */
+
+#include <fun4all/Fun4AllUtils.h>
+#include <G4_ActsGeom.C>
+#include <G4_Global.C>
+#include <G4_Magnet.C>
+#include <GlobalVariables.C>
+#include <QA.C>
+#include <Trkr_Clustering.C>
+#include <Trkr_Reco.C>
+#include <Trkr_RecoInit.C>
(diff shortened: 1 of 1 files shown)


TrackingProduction/Fun4All_TrackSeeding.C
```diff
@@ -0,0 +1,432 @@
...
+  std::string theOutfile = outfile.Data();
+  auto se = Fun4AllServer::instance();
+  se->Verbosity(1);
+
+
+  Fun4AllRunNodeInputManager *ingeo = new Fun4AllRunNodeInputManager("GeoIn");
+  ingeo->AddFile(geofile);
+  se->registerInputManager(ingeo);
+
+  CDBInterface *cdb = CDBInterface::instance();
```
osbornjd, line 118: We can remove this chunk from the macro, as I added it into tpc readout init so that it happens automatically each time.
 e.g. from line 118-130

TrackingProduction/Fun4All_TrackSeeding.C
```diff
@@ -0,0 +1,432 @@
...
+  silicon_Seeding->setinttRPhiSearchWindow(0.4);
+  silicon_Seeding->setinttZSearchWindow(1.6);
+  silicon_Seeding->seedAnalysis(false);
+  se->registerSubsystem(silicon_Seeding);
+  */
+
+  auto silicon_Seeding = new PHActsSiliconSeeding;
+  silicon_Seeding->Verbosity(0);
+  // these get us to about 83% INTT > 1
+  silicon_Seeding->setinttRPhiSearchWindow(1.0);
```
osbornjd, line 168: the rphi search window should be 0.4 and the z search window should be 2. We probably need to readjust these given the new alignment, but that is a later task

osbornjd (changes requested):

PR #1018, opened 2024-12-05
use same Micromegas matching cuts accross all macros.

hupereir:
Title says it all. 
The cuts are set to match study presented at https://indico.bnl.gov/event/25004/contributions/97983/attachments/58028/99657/tpot_matching.pdf


hupereir:
Closing this one, following Joe's comment
I'll open a new one for the ZF macro only.

PR #1020, opened 2024-12-08
Update Calo_Calib macro to work as both data and simulation calibration afterburner using timestamp

emclaughlin2:
Update Process_calo_calib function to work for both data and MC. Uses reco const TIMESTAMP variable to differentiate between data and MC so TIMESTAMP variable in reco const instance in f4a macro must be set to runnumber in order to do calibrations.


bseidlit:
Thanks!  Can you remove the stuff about the ZDC, it is not used in the data either.

bseidlit:
Some documentation 
https://indico.bnl.gov/event/25840/contributions/100350/attachments/59032/101382/12.9.24.pdf

PR #1025, opened 2024-12-17
Trigger mode/streaming mode setup added in InttUnpacker

gwd213:
Trigger mode and streaming mode has to be set before hit unpacking.
In order to use INTT triggered mode data, data taking mode is loaded from database and set the data taking mode in unpacker.


osbornjd (approved):
Re-added the tpc hit unpacking changes that were inadvertently removed

gwd213:
I found that some other subsystems unpacker has unnecessary change just due to mistake using old build on my porked repo. The issues have been addressed and change the guard pattern runnumber as 0.

PR #1030, opened 2024-12-20
Bugfix: set trigger value in jet constituent QA

ruse-traveler:
This PR fixes a minor bug: the trigger flag value wasn't being provided in the call to `ConstituentsInJets::setTrgStoSelect()`, so this module was always selecting the MBDNS >= 1 trigger.

### DIFF
--- a/JetProduction/Fun4All_JetProductionYear2.C
+++ b/JetProduction/Fun4All_JetProductionYear2.C
@@ -11,7 +11,6 @@
 #include <vector>
 
 // coresoftware headers
-#include <calotrigger/TriggerRunInfoReco.h>
 #include <ffamodules/CDBInterface.h>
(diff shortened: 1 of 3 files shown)


bseidlit:
Isn't this already in the DSTs at this stage of the prod? it is in the calo fitting macro.  If it is being striped out accidentally we should fix that.

ruse-traveler:
> Isn't TriggerRunInfoReco nodes already in the DSTs at this stage of the prod? it is in the calo fitting macro. If it is being striped out accidentally we should fix that.

Whoops! Good catch! I forgot about that! I'll double check that it isn't being stripped out. We do need the bugfix in the constituents, though, so I'll reduce the scope of this PR to just that...

ruse-traveler:
This is now descoped!

PR #1033, opened 2025-01-06
ZF macro fix - distortion

yuxdPKU:
Make some changes on ZF macro
1. add lines to turn on distortion corrections, which are off by default
2. set correct convertSeeds in TrackResidual module
3. add preliminary distortion module, which is on by default


osbornjd (commented):
Lets remove the prelim distortion correction call in the ZF seeding by default. In principle this should be a special testing situation that we probably don't want everyone running

PR #1036, opened 2025-01-14
I modified the Intt_HitUnpacking to make it able to handle different coll…

ChengWeiShih:
…ision type.

The build-in switch `TRACKING::pp_mode` is used to set the collision type. And if the `pp_mode == false`, currently we disable the time_bucket information for the `TrkrHitSets`, and we apply the bco_diff cut to reject the background hits.

### DIFF
--- a/common/Trkr_Clustering.C
+++ b/common/Trkr_Clustering.C
@@ -88,10 +88,16 @@ void Intt_HitUnpacking(const std::string& server="")
   inttunpacker->Verbosity(verbosity);
   inttunpacker->LoadHotChannelMapRemote("INTT_HotMap");
   inttunpacker->set_triggeredMode(!isStreaming); 
-   if(server.length() > 0)
-    {
-      inttunpacker->useRawHitNodeName("INTTRAWHIT_" + server);
-    }
+  if (TRACKING::pp_mode == false) // note : false -> AuAu
+  {
+    inttunpacker->writeInttEventHeader(true);
+    inttunpacker->set_bcoFilter(true); // default : false
+    inttunpacker->runInttStandalone(true); // default: false
+  }
+  if(server.length() > 0)
+  {
+    inttunpacker->useRawHitNodeName("INTTRAWHIT_" + server);
+  }
   se->registerSubsystem(inttunpacker);
 }
 void Intt_Clustering()


common/Trkr_Clustering.C
```diff
@@ -88,10 +88,16 @@ void Intt_HitUnpacking(const std::string& server="")
...
   inttunpacker->set_triggeredMode(!isStreaming); 
-   if(server.length() > 0)
-    {
-      inttunpacker->useRawHitNodeName("INTTRAWHIT_" + server);
-    }
+  if (TRACKING::pp_mode == false) // note : false -> AuAu
+  {
+    inttunpacker->writeInttEventHeader(true);
+    inttunpacker->set_bcoFilter(true); // default : false
+    inttunpacker->runInttStandalone(true); // default: false
```
osbornjd, line 95: Why should this be set in AuAu? We will have the Gl1 enabled, so how else will we determine the time bucket properly?

ChengWeiShih: In the AuAu data, currently we don't use the timebucket, which means we consider the hits of 3 crossings (trigger_bco -1, trigger_bco, trigger_bco + 1) as the hits from the same crossing. Which is actually different from the default setting of `InttCombinedRawDataDecoder`. We therefore need to change the flags.

ChengWeiShih: This is more about the hits in one strobe length (one F4A event). It's nothing to do with the GL1.

osbornjd: > In the AuAu data, currently we don't use the timebucket, which means we consider the hits of 3 crossings (trigger_bco -1, trigger_bco, trigger_bco + 1) as the hits from the same crossing

I'm not sure I understand this statement, the timebucket in all of the hitsets should then just be -1, 0, or 1. Why would we just not fill it at all? That is a good quality check to make sure that the hits look like we expect them. I don't understand how it has nothing to do with the GL1 because inherently any triggered event has something to do with the GL1

ChengWeiShih: No, the switch `runInttStandalone ` sets the `time_bucket` of all the hits into 1.

ChengWeiShih: The reason why we don't separate the hits bco by bco like what we do for pp data set is because 1. the collision date is low, 2. the coarse delay of INTT may not be fine tuned, which means there can be some hits falling into the next bco timing (trigger_bco + 1) .

ChengWeiShih: In Run24 AuAu, all the data was taken with the strobe length of 100 BCO. The events are still there, we just change the time_bucket of the hits in that triggered event.

osbornjd: If runInttStandalone sets the variable to true, then this entire block is skipped
https://github.com/sPHENIX-Collaboration/coresoftware/blob/58bba29e020d9f17b9ced5da2b3256e530dd75eb/offline/packages/intt/InttCombinedRawDataDecoder.cc#L272-L289

meaning that all time_buckets are set to 0. That means we assume that all hits are exactly in time with the GL1. I do not understand why we wouldn't use the GL1 information to actually confirm this, and then have an additional check that the event combining is properly assigning hits to the GL1 (i.e. that all hits in an event are actually assigned time bucket -1, 0, or 1). If we saw that a hit was assigned time bucket 20, or 10, or 4, or anything other than (-1, 0, 1), that is an indication that the event combining did not work as expected and thus acts as an offline QA check of the event combiner. So I'm puzzled as to why we would just assume the time bucket instead of explicitly checking it

osbornjd: Maybe I am hung up on this, but I don't understand why we would assume something that we can explicitly check and confirm

ChengWeiShih: When the trigger is fired, INTT opens the data taking window 100 BCOs for the run 54280. Therefore there can be more than one collision in one INTT triggered event. It's just the same as the pp triggered mode. The whole triggered event is matching with the GL1 regardless how long the INTT opens the data taking window (1 or 100 BCOs). And I am not sure how the GL1 can do the matching with the INTT hits 20 BCOs after the trigger BCO. Those hits are still in the same triggered event.

As an example, we can have one triggered event like 
10 hits in triggered BCO-1
300 hits in triggered BCO
30 hits in triggered BCO +1
3 htis in triggered BCO +2
10 hits in triggered BCO +3.... and more 

We can see above in the INTTRawHit already, don't need to go to the TrkrHitSet. For the Au+Au data analysis, our focus is only on the hits from the collision that fires the trigger. Therefore we can just reject the hits from most of time_buckets that are far from the triggered BCO (say we only keep the hits from trig_bco -1, trig_bco, trig_bco +1, 10 + 300 + 30 hits in the example). In addition, we kinda worry about the `INTT coarse delay fine tuning`. If the `coarse delay` is not set perfectly, there can be some hits fall into the next BCO (trig_bco +1). But those hits are actually from the `trig_bco`, so we should use them in the clustering. To do so, we need to set the `time_bucket` to be zero for all the remaining hits from that 3 BCOs. (For the run 54280, RHIC was running with 56 x 56 config. There is no collision before and after the trigger BCO)

osbornjd: I agree 100% with everything you have said. Where I disagree is that the solution is just to set the time bucket of every raw hit to 0, which is what setting runInttStandalone to true [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/58bba29e020d9f17b9ced5da2b3256e530dd75eb/offline/packages/intt/InttCombinedRawDataDecoder.cc#L272-L289) will do. We should instead have runInttStandalone = false, calculate the time bucket with respect to the GL1 (as is currently done in that code block) and then only write out the hits that correspond to time bucket -1, 0, or 1 in triggered mode. At the moment, if you just set runInttStandalone = true, then every hit from your example will be assigned to the triggered BCO which is wrong. i.e. 10 hits in triggered BCO -1 will have time bucket 0, 300 hits in triggered BCO will have time bucket 0, 3 hits in triggered BCO+2 will have time bucket 0, etc. That is not the behavior we want to have

ChengWeiShih: For the p+p, what you say is correct. We don't want to mix up the hits with different time bucket, as they can be from different collisions, therefore we should not merge them. 

But for the AuAu run 54280, the number of crossings is 56, which means, there is no collision one-bco before and after the trigger BCO, so even if we merge the `hits from trig_bco-1`, `hits from trig_bco` and `hits from trig_bco+1`, there can still be one collision in this three-bco time window.

Sure, we can do what you say in AuAu data, just the same way as how we process the hits for the pp runs.  But the caveat is if there are some hits with incorrect hit timing assignments (hit timing falling to the next bco (trig_bco+1)), we might introduce more clusters than what it should be. Like the example in the following. 

Assuming there are 310 hits in the trigger bco (trig_bco), due to the imperfect `INTT coarse delay setting`, some hits from the `trig_bco` actually end up with having the hit timing of (trig_bco + 1) (incorrect hit timing assignment). 
Therefore what we see in the data would be like: 
300 hits in trig_bco -> (310 - 10, as there are 10 hits with incorrect hit timing assignment)
12 hits in trig_bco + 1 -> (10 hits from the trig_bco & assume 2 noise hits in trig_bco + 1)

if we simply don't care about the `incorrect hit timing assignment`, and we do the clustering time_bucket by time_bucket, we will end up with having more clusters than what it should be. (say there can be one cluster made of two hits, but if one of the hits are with the hit timing of trig_bco+1, this cluster will be segmented into two clusters due to the incorrect hit timing assignment.)

common/Trkr_Clustering.C
```diff
@@ -88,10 +88,16 @@ void Intt_HitUnpacking(const std::string& server="")
   inttunpacker->Verbosity(verbosity);
   inttunpacker->LoadHotChannelMapRemote("INTT_HotMap");
   inttunpacker->set_triggeredMode(!isStreaming);
```
osbornjd, line 90: It also occured to me that I was wrong about using the `pp_mode` flag, we should really just use the result of the database query which sets triggered mode here. So in fact, what is different about the needed flags that are set in your PR that is not already addressed by the `set_triggeredMode` flag?

ChengWeiShih: The modification here is mainly for accounting for the `incorrect hit timing assignment`, as explained above.

ChengWeiShih: The trigger mode flag is for the time_bucket peak offset.

osbornjd: Right, can we group everything together so that there is only one flag that sets whether or not the hit unpacker is triggered or streaming, and that function then does all of the necessary setup?

common/Trkr_Clustering.C
osbornjd: If we set all the hits time bucket to 0, doesn't that introduce unnecessary background then? You are basically trading off the hits that were incorrectly assigned to time bucket GL1 BCO + 1 (that should have been in GL1 BCO) at the expense of adding in all hits from all BCOs not associated to the GL1. So then you will cluster in a single time bucket, but your background will be much higher?

common/Trkr_Clustering.C
osbornjd: My point is, the issue being discussed is not a pp vs Au Au issue, it is really a triggered vs streaming mode issue, right?

common/Trkr_Clustering.C
ChengWeiShih: Yes, you are right, this is the trade off, to have all the signal hits considered in the clustering, the noise hits are also included. I am not sure whether we can assign the syst. unc. for this by doing the clustering w/ and w/o the time_bucket.

common/Trkr_Clustering.C
ChengWeiShih: I think it can only happen in Au+Au, I meant, for the p+p trigger mode, you 100% don't want to mix-up the hits with different crossings (due to the polarities of the bunches.)

common/Trkr_Clustering.C
osbornjd: Okay, I would suggest preparing a set of slides that describes this issue to discuss in the tracking meeting either tomorrow or Friday. I think we should discuss the issue as a group to decide what is the best way forward - has this been already discussed in an INTT meeting?

ChengWeiShih: Sorry for the late reply. I was in the TPS annual meeting. Yes, I think I can give a presentation in the tracking meeting. Say Monday? This has been discussed in the INTT meeting. We then came up with such a way to handle the AuAu clustering.

PR #1059, opened 2025-02-07
Increased TPOT timing window. 

hupereir:
This should be enough to accomodate extended readout hits.
I will show a few plots at the next tracking meeting.


osbornjd:
Merging now so that it is in next weeks ana build

hupereir:
> Merging now so that it is in next weeks ana build

Thanks !

PR #1064, opened 2025-02-15
Implement TPC time zero offset via alignment transforms in macros

adfrawley:
Accompanies coresoftware PR #3438, should not be merged until after that.

Enables the application of the TPC time zero offset using alignment transforms.

### DIFF
--- a/TrackingProduction/Fun4All_TrackAnalysis.C
+++ b/TrackingProduction/Fun4All_TrackAnalysis.C
@@ -97,6 +97,10 @@ void Fun4All_TrackAnalysis(
   se->registerInputManager(ingeo);
 
   TpcReadoutInit( runnumber );
+  // these lines show how to override the drift velocity and time offset values set in TpcReadoutInit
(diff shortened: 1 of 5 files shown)


osbornjd (commented):
Is there a reason to have a separate CDB file for the t0? e.g. can we just have a single CDB file that contains both the drift velocity and t0, load that in once, and then read the float values from the file for both?

pinkenburg:
When we came up with this the problem was that the t0 and drift velocity change independently (means different validity ranges) which made keeping track of this difficult. If we are at a stage where both change run by run, then we should lump them together

adfrawley:
No reason, other than a version of the drift velocity exists, but the t0 does not (yet).  We would need a mechanism to detect when one of the pair does not exist in the CDB. Otherwise there would be no way to know that you are using a value for one that is not calibrated.

osbornjd:
Chris makes a good point, and principle of the t0 should really be only run by run-dependent but kristoff has mentioned before that we will probably want drift velocity calibrations that very even within a run. So it's probably a better idea to keep these separate

PR #1073, opened 2025-02-26
Simplified/removed calls following intt library changes

josephbertaux:
* PHG4InttDeadMapLoader is now moot

### DIFF
--- a/common/G4_TrkrSimulation.C
+++ b/common/G4_TrkrSimulation.C
@@ -11,7 +11,6 @@
 #include <g4mvtx/PHG4MvtxHitReco.h>
 #include <g4mvtx/PHG4MvtxSubsystem.h>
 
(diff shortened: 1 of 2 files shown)


josephbertaux:
I wanted to ask about the two failing checks. I'm guessing the gcc one hasn't been re-run b/c the our minor version has changed, and there is a corresponding one that passed.

The one I'm concerned about is the tracking reconstruction check. It hasn't been re-run, I mostly wanted to verify that this results in no change in that output. I can probably run it locally with some hand-holding.

osbornjd:
They might be benign. Let's just run another build and see what happens overnight

osbornjd:
We don't have data QA integrated into jenkins yet. But the simulated QA all looks the same. To test the changes to the decoder you'd have to just run a test over a few events with and without the change to see if there is any appreciable difference to the underlying hit distribution

PR #1074, opened 2025-02-27
fix static distortion flag for data

yuxdPKU:
``G4TPC::DISTORTIONS_USE_PHI_AS_RADIANS`` is for simulation
``G4TPC::USE_PHI_AS_RAD_STATIC_CORRECTIONS`` is for data


osbornjd:
Does this mean we have not been applying the static distortions correctly in our studies?

yuxdPKU:
Maybe not, because the default value is also false
```c++
common/G4_TrkrVariables.C:  bool USE_PHI_AS_RAD_STATIC_CORRECTIONS = false;
```
But I think fix this typo will reduce some confusion.

PR #1095, opened 2025-04-05
NoBkgdSubJetReco.C update

dadingsphenix:
make sure all the jets making from constituents functions have fastjet algorithm generated


dadingsphenix:
Make sure all algo are declared and defined properly

pinkenburg:
@dadingsphenix you need to resolve the merging conflict. It looks like an easy one and you can use the resolve conflicts button.

dadingsphenix:
@pinkenburg Thank you! I think it is all fixed now

PR #1099, opened 2025-04-24
set MVTX_APPLYMISALIGNMENT to true

pinkenburg:
Finally make this the new default

### DIFF
--- a/common/G4_TrkrVariables.C
+++ b/common/G4_TrkrVariables.C
@@ -13,7 +13,7 @@ namespace Enable
   bool MVTX_CLUSTER = false;
   bool MVTX_QA = false;
   bool MVTX_SUPPORT = false;
-  bool MVTX_APPLYMISALIGNMENT = false;
+  bool MVTX_APPLYMISALIGNMENT = true;
 
   int MVTX_VERBOSITY = 0;
 


osbornjd:
I don't object to the point of this PR, but just to comment this is likely to wreak havoc on the QA. I started a build to see what the effect is

osbornjd:
Good thing we ran the CI... a couple notes from the log file https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/7112/console

1.  The alignment isn't actually loaded. I'm guessing the proper file isn't in the simulated CDB tag, it is loaded from https://github.com/sPHENIX-Collaboration/coresoftware/blob/f61adf1d092dddbc7b87e853e2b2c092ae33f5eb/offline/packages/trackreco/MakeActsGeometry.cc#L1046
2. There is a mismatch between the Acts geometry and the g4 geometry for some reason. I'm looking into why

osbornjd:
Actually on second thought, there might be a mismatch because the alignment isn't properly loaded into G4 from the CDB but Acts thinks that there should be... So fixing the CDB tag might solve both issues

osbornjd:
This still dies for the same reason, although the log indicates that the misaligned geometry is now correctly being loaded. So now something else is apparently going on

PR #1104, opened 2025-04-28
Comment out MVTX hit pruner

hupereir:
- comment out MVTX hit pruner
- also load pedestal file in Micromegas clusterizer

Removing the MVTX hit pruner module makes the code consistent with official production code, as discussed with @jdosbo

### DIFF
--- a/common/Trkr_Clustering.C
+++ b/common/Trkr_Clustering.C
@@ -60,10 +60,10 @@ void Mvtx_Clustering()
   int verbosity = std::max(Enable::VERBOSITY, Enable::MVTX_VERBOSITY);
   Fun4AllServer* se = Fun4AllServer::instance();
 
-  // prune the extra MVTX hits due to multiple strobes per hit
-  MvtxHitPruner* mvtxhitpruner = new MvtxHitPruner();
-  mvtxhitpruner->Verbosity(verbosity);
-  se->registerSubsystem(mvtxhitpruner);
+//   // prune the extra MVTX hits due to multiple strobes per hit
+//   auto mvtxhitpruner = new MvtxHitPruner;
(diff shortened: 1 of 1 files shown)


osbornjd (commented):
This hurts our matching efficiency pretty badly, although we need to do it. I think we need to understand first why the seeder downstream does not just handle this change easily (since it in principle should, given that it sorts things on a strobe by strobe basis)

hupereir:
Hi Joe, thanks for the feedback, this is very instructive, and raise several questions: 
- if this hurts the matching on simulations, should we not expect it to hurt it also on real data, and then should it be re-added there ? (at least temporarily) 
- ultimately I think 'some' version of the pruner should be useful. Rather than merging all hits on the same strobes we could look for duplicated hits (or in a more robust manner, for duplicated clusters) accross adjacent probes, and discards such duplicates. You should get the best of both worlds. This goes along the line of what I am working on now with duplicated seeds. 
Lets not commit this change for now, then :)
I'll mark it as draft.

hupereir:
@osbornjd @adfrawley 

The reason why removing the MvtxHitPruner in MC breaks the MVTX matching efficiency is because the default strobe window in PHActsSiliconSeeder is too narrow, even for simulations. 

Our simulations generates hits in strobes 0, 1 and 2, and all hits in strobe 2 were discarded, unless one runs the hitpruner. (which moved them back to strobe 0). 

Extending the strobe range (in the Trkr_Reco macro) fixes the issue here. 
@osbornjd can you double check and merge if true ? 

Being able to remove the HitPruner will make the code more similar to what we run on real data.

hupereir:
Screenshot_20250520_155022

PR #1107, opened 2025-04-30
add switches to distinguish calos and tracks, raw and subtracted inputs

dadingsphenix:
add switches to add only towers or tracks based on the need; add switches to select background subtracted/not-background subtracted inputs, and turn on/off modules as needed

### DIFF
--- a/common/Jet_QA.C
+++ b/common/Jet_QA.C
@@ -41,7 +41,9 @@ R__LOAD_LIBRARY(libjetqa.so)
 	//! Set to true if input jets utilize calorimeters(e.g. via particle flow)
 	bool HasCalos = false;
 
-	bool UseBkgdSub = false;//swicth under HasCalos
+	//! Set to true if jets have background subtracted
+        //! (e.g. HIJetReco.C output)
+	bool UseBkgdSub = false;
 
 	//! Set to true to generate histograms for no trigger selection
 	bool DoInclusive = true;
@@ -379,21 +381,18 @@ R__LOAD_LIBRARY(libjetqa.so)
                     Type::AntiKtTowerR04,
                     Type::AntiKtTowerR05
                   });
-                }
-		  	     
+                }  	     
   	     	break;
 
 		case Source::Tracks:
-	     	if (HasTracks)
-		{	
-	          vecJetsToQA.insert(vecJetsToQA.end(),
-                  {
+
+	        vecJetsToQA.insert(vecJetsToQA.end(),
+                {
                   Type::AntiKtTrackR02,
                   Type::AntiKtTrackR03,
                   Type::AntiKtTrackR04,
                   Type::AntiKtTrackR05
-                  });	
-	        }
+                });	
 		break;
 
              }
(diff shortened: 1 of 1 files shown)


common/Jet_QA.C
```diff
@@ -40,6 +40,10 @@ R__LOAD_LIBRARY(libjetqa.so)
 	//! Set to true if input jets utilize calorimeters(e.g. via particle flow)
 	bool HasCalos = false;
 
+	bool QA_SubtractedTowers = false;//swicth under HasCalos
+
+	bool QA_RawTowers = false;//switch under HasCalos
```
ruse-traveler, line 45: We might not need this switch since (at least in production) there's never going to be a case where we run both the background-subtracted and no-background-subtracted jet reconstruction in the same process. Then if a user really wants to run both, they can get both QA versions by doing something like this in their macro:
```
  // register bkgd-sub jet qa modules
  JetQA::UseBkgdSub = true;
  Jet_QA();

  // register no-subtraction jet qa modules
  JetQA::UseBkgdSub = false;
  Jet_QA();
```

That means we can rely on `!UseBkgdSub` to get the no-background-subtracted equivalents!
```suggestion
```

common/Jet_QA.C
```diff
@@ -344,22 +348,43 @@ R__LOAD_LIBRARY(libjetqa.so)
 	// --------------------------------------------------------------------------
 	inline std::vector<uint32_t> GetJetsToQA()
 	{
-
-		std::vector<uint32_t> vecJetsToQA = {
+	     std::vector<uint32_t> vecJetsToQA;
+	     if (HasCalos)
+	     {
+		   if (QA_SubtractedTowers)
```
ruse-traveler, line 354: In line with above suggestion:
```suggestion
		   if (UseBkgdSub)
```

common/Jet_QA.C
```diff
@@ -344,22 +348,43 @@ R__LOAD_LIBRARY(libjetqa.so)
...
 			Type::AntiKtTowerSubR03,
 			Type::AntiKtTowerSubR04,
-			Type::AntiKtTowerSubR05,
-			Type::AntiKtTowerR02,
+			Type::AntiKtTowerSubR05
+		     });
+		   }
+
+		   if (QA_RawTowers)
+                   {
```
ruse-traveler, line 366: In line with above comment:
```suggestion
		   }
		   else
		   {
```

common/Jet_QA.C
```diff
@@ -344,22 +348,43 @@ R__LOAD_LIBRARY(libjetqa.so)
...
 	inline std::vector<uint32_t> GetJetsToQA()
 	{
-
-		std::vector<uint32_t> vecJetsToQA = {
+	     std::vector<uint32_t> vecJetsToQA;
+	     if (HasCalos)
+	     {
+		   if (QA_SubtractedTowers)
+		   {	   
+                     vecJetsToQA.insert(vecJetsToQA.end(),
```
ruse-traveler, line 356: This is heading in the right direction! But I don't think this is quite what we need....

For example, suppose we're running with both track-jets and calo-jets: in the `JetsWithTrksQA` function, we'll still want to process only track jets, but when we call `std::vector<uint32_t> vecJetsToQA = JetQA::GetJetsToQA();`, we'll _also_ get the calo jets! And you'll get an error when Fun4All tries to run a track-jet module on calo jets...

My suggestion would be to use something like a switch case:
```
std::vector<uint32_t> GetJetsToQA(const uint32_t source)
{
  //... etc. ...//
  switch (source)
  {
    case 0:
      vecJetsToQA.insert(
        vecJetsToQA.end(),
         {
           //... load the jets here ...//
          }
        );
        break;
      //... etc ...//
    } 
   //... etc ...//
}
```
And then, naturally, you can create a new enum to give the cases nice names to the sources (e.g. `Source::Tracks`)

common/Jet_QA.C
```diff
@@ -393,90 +418,108 @@ void CommonJetQA(std::optional<uint32_t> trg = std::nullopt)
...
-	}
-	se -> registerSubsystem(kinematicQA);
+	  if (JetQA::QA_SubtractedTowers)
+	  {	  
+	    JetKinematicCheck* kinematicQA = new JetKinematicCheck(
+	    "JetKinematicCheck" + trig_tag + "_towersub1_antikt",
+	    JetQA::JetInput[JetQA::Type::AntiKtTowerSubR02],
+	    JetQA::JetInput[JetQA::Type::AntiKtTowerSubR03],
+	    JetQA::JetInput[JetQA::Type::AntiKtTowerSubR04],
+	    JetQA::JetInput[JetQA::Type::AntiKtTowerSubR05]);
```
ruse-traveler, line 431: Then if you go with the switch-case approach, you can `GetJetsToQA` to get the appropriate jets to QA here regardless of which mode you're in.
```suggestion
	    "JetKinematicCheck" + trig_tag + "_towersub1_antikt",
	    JetQA::JetInput[vecJetsToQA[0]],
	    JetQA::JetInput[vecJetsToQA[1]],
	    JetQA::JetInput[vecJetsToQA[2]],
	    JetQA::JetInput[vecJetsToQA[3]]);
```
(And then, another enum to index the resolutions -- e.g. `Res::R03` -- can help with readibility)

ruse-traveler (commented):
Thanks, Dading! This looks great! I have a couple suggestions for how to handle switching between sources to consider (see comments below).

common/Jet_QA.C
```diff
@@ -385,98 +422,117 @@ void CommonJetQA(std::optional<uint32_t> trg = std::nullopt)
...
+	std::vector<uint32_t> vecJetsToQA;
+   	if (JetQA::HasCalos)
+   	{
+     	  vecJetsToQA = JetQA::GetJetsToQA(JetQA::Source::Calos);
+     	  if (JetQA::HasTracks)
+          {
+            auto tmp = JetQA::GetJetsToQA(JetQA::Source::Tracks);
+            vecJetsToQA.insert(vecJetsToQA.end(), tmp.begin(), tmp.end());
+          }
+        }
```
ruse-traveler, line 497: Careful! If you're running with only tracks, this block will never execute! Instead you could do something like so:
```suggestion
   	if (JetQA::HasCalos)
   	{
   	  auto vecCalJets = JetQA::GetJetsToQA(JetQA::Source::Calos);
   	  vecJetsToQA.insert(vecJetsToQA.end(), vecCalJets.begin(), vecCalJets.end());
        }
     	if (JetQA::HasTracks)
        {
          auto vecTrkJets = JetQA::GetJetsToQA(JetQA::Source::Tracks);
          vecJetsToQA.insert(vecJetsToQA.end(), vecTrkJets.begin(), vecTrkJets.end());
        }
```

common/Jet_QA.C
```diff
@@ -511,43 +567,40 @@ void JetsWithTracksQA(std::optional<uint32_t> trg = std::nullopt)
...
+	    .verbose     = Enable::QA_VERBOSITY,
+	    .doDebug     = false,
+	    .doInclusive = false,
+	    .doInJet     = true,
+	    .doHitQA     = false,
+	    .doClustQA   = true,
+	    .doTrackQA   = true,
+	    .doJetQA     = true,
+	    .rJet        = 0.4,
+	    .jetInNode   = JetQA::JetInput[jet]
```
ruse-traveler, line 597: Minor suggestion in line with indenting comment above.
```suggestion
	      .outMode     = TrksInJetQA::OutMode::QA,
	      .verbose     = Enable::QA_VERBOSITY,
	      .doDebug     = false,
	      .doInclusive = false,
	      .doInJet     = true,
	      .doHitQA     = false,
	      .doClustQA   = true,
	      .doTrackQA   = true,
	      .doJetQA     = true,
	      .rJet        = 0.4,
	      .jetInNode   = JetQA::JetInput[jet]
```

ruse-traveler (commented):
Almost there! I think there are few redundant checks on `HasTrks`/`HasCalos` that can be removed to simplify the code, and there's a bug in the selection of jets before the R-dependent common jet modules.

Aside from a few minor formatting suggestions, that should be it!

ruse-traveler (approved):
Awesome! Looks great!

PR #1110, opened 2025-05-03
remove tpc clustering min error thresholds

devloom:
Tracking simulation performance in isobutane can be restored by removing the minimum error cuts on clusters.

### DIFF
--- a/calibrations/calo/calo_cdb/scripts/runProd.py
+++ b/calibrations/calo/calo_cdb/scripts/runProd.py
@@ -4,13 +4,15 @@ This module generates a list of run / datasets given a run type and event thresh
 """
 import argparse
 import datetime
-import glob
 import logging
 import os
 import shutil
 import subprocess
 import sys
+import textwrap
 
+from pathlib import Path
+from multiprocessing import Pool, cpu_count
 import pandas as pd
 from sqlalchemy import create_engine, text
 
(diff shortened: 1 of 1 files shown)


osbornjd:
Here's a build, let's check the QA to make sure it reproduces what @devloom was able to produce locally.
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc14/734/

devloom:
@osbornjd @bogui56  This PR goes along with https://github.com/sPHENIX-Collaboration/coresoftware/pull/3784 and should hopefully allow for the use of the same clusterizer settings as raw data by implementing pedestal subtraction and zero suppression in the PHG4TpcDigitizer.

osbornjd:
I setup a Jenkins build using both this and the coresoftware PR, it will show up attached to this PR

common/G4_TrkrSimulation.C
```diff
@@ -602,8 +602,12 @@ void TPC_Cells()
   digitpc->SetTpcMinLayer(G4MVTX::n_maps_layer + G4INTT::n_intt_layer);
   double ENC = 670.0;  // standard
   digitpc->SetENC(ENC);
+  double pedestal = 40540; // 40540 electrons = 60 ADU
```
osbornjd, line 605: We should consider how we actually store these values - might be something we want to put in the CDB

devloom: That works for me. Are the pedestal and zero suppression values for the raw data unpacking stored in the CDB already? If so, I can adjust the code so it is compatible with ADU values instead of # electrons.

PR #1125, opened 2025-05-24
alternative fitting macro for the seb by seb combining workflow

Shuonli:
Calo fitting macro for the new workflow where we read in the combined DSTs from different seb machines from different DST files.


pinkenburg:
I just checked in a slightly different macro which does a better job extracting the run number/segment and set the output filenames accordingly. Closing this PR

PR #1127, opened 2025-05-25
new magnet macro

Shuonli:
use new BaBar Geant code based magnet geom as default.

switch `MAGNET_OLD` to true to use the old geom

### DIFF
--- a/common/G4_Magnet.C
+++ b/common/G4_Magnet.C
@@ -5,6 +5,8 @@
 
 #include <g4detectors/PHG4CylinderSubsystem.h>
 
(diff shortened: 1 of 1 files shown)


Shuonli:
convert this to draft, since merging it will might cause Jenkins complaint about other PR(as the G4 geometry will be changed so we need to update the Jenkins reference).

Shuonli:
This is due to the missing parameter initialization used in the stepping action, this PR(https://github.com/sPHENIX-Collaboration/coresoftware/pull/3612) would fix it.

PR #1137, opened 2025-06-04
added MvtxClusterPruner, as a replacement to the removed MvtxHitPruner.

hupereir:
This macro adds the MvtxClusterPruner in the Mvtx clustering part of the macro, as a replacement for the now removed MvtxHitPruner. 

I'll check the impact on the tracking QA when available, before merging.

### DIFF
--- a/common/Trkr_Clustering.C
+++ b/common/Trkr_Clustering.C
@@ -15,7 +15,7 @@
 
 #include <intt/InttClusterizer.h>
 #include <mvtx/MvtxClusterizer.h>
-#include <mvtx/MvtxHitPruner.h>
+#include <mvtx/MvtxClusterPruner.h>
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wundefined-internal"
@@ -55,17 +55,23 @@ void Mvtx_HitUnpacking(const std::string& felix="")
     }
   se->registerSubsystem(mvtxunpacker);
 }
+
 void Mvtx_Clustering()
 {
   int verbosity = std::max(Enable::VERBOSITY, Enable::MVTX_VERBOSITY);
   auto se = Fun4AllServer::instance();
 
-  // For the Mvtx layers
-  //================
-  auto mvtxclusterizer = new MvtxClusterizer("MvtxClusterizer");
+  // clusterizer
+  auto mvtxclusterizer = new MvtxClusterizer;
   mvtxclusterizer->Verbosity(verbosity);
   se->registerSubsystem(mvtxclusterizer);
+
+  // cluster pruner
+  auto mvtxClusterPruner = new MvtxClusterPruner;
+  mvtxClusterPruner->set_use_strict_matching(true);
+  se->registerSubsystem(mvtxClusterPruner);
 }
+
 void Intt_HitUnpacking(const std::string& server="")
 {
   int verbosity = std::max(Enable::VERBOSITY, Enable::INTT_VERBOSITY);


hupereir:
Screenshot_20250604_151129
Looking at the QA output, the main difference I see is in the lefthand plot of n_cluster/track. 
This is only for the _gen histogram, not for the _reco histogram, which is identical to the reference. 
I do not know how to interpret that. Especially since the other histograms (e.g. tracking efficiency) are unchanged. 
Any insight ?

hupereir:
Also, this might have already been in previous pull requests (looking at the P values), but I cannot check right now due to: 
Screenshot_20250604_151658

osbornjd:
Unfortunately I can't see other QA pages because of the limit exceeded problem you already found out about from Jin. If this is present in other PRs then it must have snuck in before, but I don't know from what PR (and can't go check because the rate limit has been exceeded...). I will try in the morning when we presumably are starting fresh

hupereir:
@osbornjd I'll try to find when this change occured tomorrow morning. I also noticed some weirdness in the MVTX QA (also pre-dating this change): 
Screenshot_20250604_203536

hupereir:
(it is interesting that the KS-test is nearly one, while the scale is obviously wrong. )

hupereir:
@osbornjd 
https://github.com/sPHENIX-Collaboration/coresoftware/pull/3621 is the first pool request that shows the change in #clusters per truth track, as far as I can tell. 
But I don't think it has anything to do with it. 
One should look for a coincidental commit in the macros directory instead. 
Could that be the removal of the MvtxHitPruner again ? 
I'll keep looking tomorrow.

osbornjd:
Here is the master build from May 29th 3:35 AM which shows the changes
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/sPHENIX_CoreSoftware_PullRequest/9140/
Very strangely, the MVTX QA histograms have p values <1 but the "summary" p value report is listed as 1 in this build...

osbornjd:
And last bit of sleuthing, here is a build that started just ~1 hour after the hit pruner merge, which shows the changes
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/sPHENIX_CoreSoftware_PullRequest/9135/

The next previous successful build with all p values of 1 (and inspecting the histos, looks like they should be) is 
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/sPHENIX_CoreSoftware_PullRequest/9133/
which was ~4 hours prior to the hit pruner merge

hupereir:
I see, I'll look into it. My impression is that starting builds from 
Screenshot_20250605_064729
does not automatically incorporate the macro changes that we try to test, and only the next build with the macros pull request accepted does. Or does it ? Maybe I misunderstood how this is expected to work. 
If this is the case we have no real way to test macro changes with Jenkins before incorporating the changed ... or at least I don't understand the workflow.
We should clarify this with @pinkenburg and @blackcathj 
In the meanwhile I'll mark this one as draft and investigate the changes due to commenting out the hit prunner.

osbornjd:
> does not automatically incorporate the macro changes that we try to test, and only the next build with the macros pull request accepted does. Or does it ? Maybe I misunderstood how this is expected to work.

If you start a macro build from that button, it should automatically include the macro PR. For example, if I click it here then you see that the macros area of the build is already setup with the PR information.

Screenshot 2025-06-05 at 9 00 11 AM

osbornjd:
The p values for the combined build are all now 1
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/7334/
unfortunately I can't actually access the histograms because of the API limit being reached again... so I think we should just merge these and re-confirm next time we can access them

PR #1140, opened 2025-06-05
Macro for running over AuAu raw hits to prepare for Run25

aopatton:


TrackingProduction/Fun4All_raw_hit_KFP.C
```diff
@@ -0,0 +1,713 @@
...
+ TRACKING::pp_mode = false;
+
+  // distortion calibration mode
+  /*
+   * set to true to enable residuals in the TPC with
+   * TPC clusters not participating to the ACTS track fit
+   */
+  G4TRACKING::SC_CALIBMODE = false;
+
+  ACTSGEOM::mvtxMisalignment = 100;
```
osbornjd, line 179: We don't need these misalignmetn factors anymore

TrackingProduction/Fun4All_raw_hit_KFP.C
```diff
@@ -0,0 +1,713 @@
...
+  if (file.good())
+  {
+    string moveOutput = "mv " + full_file_name + " " + final_path;
+    system(moveOutput.c_str());
+  }
+}
+
+void Fun4All_raw_hit_KFP(
+    const int nEvents = 100,
+    const int runnumber = 53871,
```
osbornjd, line 125: The run number and segment are extracted from the filenames in line 271, we don't need to pass these as arguments

TrackingProduction/Fun4All_raw_hit_KFP.C
```diff
@@ -0,0 +1,713 @@
...
+
+  /*
+   * Begin Track Seeding
+   */
+
+  /*
+   * Silicon Seeding
+   */
+
+  /*
```
osbornjd, line 332: Let's just remove this, it is old from commissioning days and not needed

TrackingProduction/Fun4All_raw_hit_KFP.C
```diff
@@ -0,0 +1,713 @@
...
+    se->registerSubsystem(tpcqa);
+
+  }
+
+  //KFParticle dependancy
+     Global_Reco();
+
+  //KFParticle setup
+
+  KFParticle_sPHENIX *kfparticle = new KFParticle_sPHENIX("pipi_reco");
```
osbornjd, line 586: Can we add a flag to turn on/off the KFP stuff, as initially we will need to do some tracking studies and many won't need KFP to be run

osbornjd (changes requested):
Can you put this in the run3auau directory under TrackingProduction? We are going to have to start keeping our run3 and run2 macros separate due to the changes in the DAQ

PR #1155, opened 2025-06-25
Add Calo/Track-Specific Jet Production Macros

ruse-traveler:
This PR back-propagates the calo/track-specific jet production macros introduced in [ProdFlow#115](https://github.com/sPHENIX-Collaboration/ProdFlow/pull/115).

### DIFF
--- a/JetProduction/Fun4All_CaloJetProductionYear2.C
+++ b/JetProduction/Fun4All_CaloJetProductionYear2.C
@@ -1,16 +1,20 @@
 #ifndef FUN4ALL_CALOJETPRODUCTIONYEAR2_C
 #define FUN4ALL_CALOJETPRODUCTIONYEAR2_C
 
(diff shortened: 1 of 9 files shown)


pinkenburg:
The order of the include files has to change.
the first one has to be 
#include <GlobalVariables.C>
followed by an empty line so clang-format does not change the order. 
Then our *.C macro includes
Then our includes like
#include <mbd/MbdReco.h>
Then root include (do not see any here)
Then system includes like
#include <iostream>

ruse-traveler:
> The order of the include files has to change. the first one has to be #include <GlobalVariables.C> followed by an empty line so clang-format does not change the order. Then our *.C macro includes Then our includes like #include <mbd/MbdReco.h> Then root include (do not see any here) Then system includes like #include

Gotcha! Just pushed a commit to change the header order!

PR #1173, opened 2025-08-12
Revert Use of isSim for the CaloTowerStatus

Steepspace:
This reverts commit ab182f9cea2a1c1b7907b51788f7cfcb2a86f54c and f9c6353dd8a3eb3098b09161f71194d4ef7912dc.


pinkenburg:
lets merge this - I think the jenkins failures are caused by the old macro which calls removed methods

PR #1176, opened 2025-08-15
Added sEPD support to Calo_Calib + sEPD geometry node creation in fitting macro

largewedge:
See https://github.com/sPHENIX-Collaboration/coresoftware/pull/3828 . It also may make more sense to use sEPDGeomMapping in Fun4All_CaloProduction.C to match other calos.


bseidlit:
The changes to CaloTowerCalib look good to me, even this time saving continue statement, nice!
It will take me some time to review the rest of the code though.  I don't think we want to make any changes that would effect the event plane in the 2024 auau data until after the Initial Stages push.   In the meantime, if there aspects of the calibration code that we want to get into coresoftware (that doesn't effect the core functionality of the EPD), that would be super beneficial use of this time.

PR #1215, opened 2025-10-07
Change sampa bias correctio to match new sampa response simulation

adfrawley:
The new sampa response simulation requires an update to the sampa bias correction in the reconstruction chain.
The new value of -65 ns works moderately well with Hugo's correction to the existing drift velocities. A little additional fine tuning of the drift velocity produces good matching to the silicon in data for all crossings.


sPHENIX-GitHub-l:
I should have mentioned that this macros PR is to go with coresoftware PR 3922.

Tony

PR #1345, opened 2026-06-23
chore: pp mode to streaming mode

osbornjd:
The names `pp_extended_readout_time` and `pp_mode` had caused confusion about whether or not they apply to the OO data. The streaming readout reconstruction is the same, so this PR refactors the variables to a more general name

**Motivation / context**

This PR refactors tracking configuration flag names to use more general *streaming_mode* terminology, avoiding ambiguity about whether the settings apply only to pp data. The intent is to keep reconstruction behavior consistent while improving clarity.

**Key changes**
- Replaced `TRACKING::pp_mode` with `TRACKING::streaming_mode` across jet production, tracking production, calibration, and common reconstruction macros.
- Renamed `TRACKING::pp_extended_readout_time` to `TRACKING::extended_readout_time`, and updated pileup/readout timing logic and diagnostics to use the new names.
- Updated downstream configuration plumbing to pass `streaming_mode` through relevant `set_pp_mode(...)` APIs and conditionals (including `embed_scan` gating and key tracking steps like vertexing/matching/fitting/cleaning).
- Updated `common/GlobalVariables.C` to remove the old `pp_*` globals and introduce `streaming_mode` / `extended_readout_time`.
- Residual references to the old `pp_mode` / `pp_extended_readout_time` identifiers appear to be limited to commented-out lines (no active logic found in the quick scan).

**Potential risk areas**
- Reconstruction behavior changes if any component interprets `streaming_mode` differently than the prior `pp_mode` semantics.
- Compatibility risk for external workflows/macros that may still reference the old global names (even if this repo is updated).
- Timing/pileup effects if `extended_readout_time` is consumed differently than `pp_extended_readout_time`.
- Thread-safety concerns are unlikely to increase from a rename alone, but should be re-validated where these globals are read during initialization.

**Possible future improvements**
- Remove/clean remaining commented references to the old `pp_*` terminology and update any documentation/examples.
- Add/confirm regression checks (e.g., tracking QA and timing/pileup validation) to verify behavior is unchanged end-to-end.
- If external users need time to migrate, consider a deprecation/compatibility layer for the old names.

AI can make mistakes, so please use best judgment when reviewing the renamed configuration paths and their downstream effects.

### DIFF
--- a/JetProduction/Fun4All_CaloJetProductionYear2.C
+++ b/JetProduction/Fun4All_CaloJetProductionYear2.C
@@ -107,7 +107,7 @@ void Fun4All_CaloJetProductionYear2(
   G4TPC::ENABLE_MODULE_EDGE_CORRECTIONS = true;
   Enable::MVTX_APPLYMISALIGNMENT        = true;
   ACTSGEOM::mvtx_applymisalignment      = Enable::MVTX_APPLYMISALIGNMENT;
-  TRACKING::pp_mode                     = NSJETS::is_pp;
+  TRACKING::streaming_mode                     = NSJETS::is_pp;
 
   // initialize interfaces, register inputs -----------------------------------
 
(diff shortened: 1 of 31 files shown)


osbornjd:
I will wait to merge this until next week so that we can discuss it in a tracking meeting and give people a fair warning

PR #1346, opened 2026-06-23
chore: run tpc-tpot by default

osbornjd:
As discussed in the tracking meeting with @hupereir, we should have this run in the default production and then just ignore the TPOT clusters in the fitting for now. This puts the TPC TPOT matching back into the default workflow.


hupereir:
Hi Joe, thanks for re-enabling track to TPOT matching ! 
Now I think you also want to remove TPOT from the trackfit though. 
Typically "actsFit->setUseMicromegas(false);"
Right now it is set to actsFit->setUseMicromegas(G4TRACKING::SC_USE_MICROMEGAS);
(which is somewhat ill-named ... ) Still, I leave it to you to either disable it permanently or change the value of SC_USE_MICROMEGAS to false in the relevant calling macro.

osbornjd:
Thanks Hugo, okay done. Merging this now

PR #1359, opened 2026-07-14
Add run38

pinkenburg:
This PR applies the changes for the run 38 wide vertex sims


pinkenburg:
vertex distribution for wide setting:
g4vtx_wide_vtx

PR #1362, opened 2026-07-17
Make the rescaling of outlier calib. factors harmless

asasikum:
Changed the condition for resetting the calibration factors to a global average, to avoid rescaling of real hot or cold towers. (noticed this bug during hcal calibration of run3OO dataset)

## Motivation / context

During HCAL calibration of the run3OO dataset, outlier calibration-factor rescaling could incorrectly modify genuine hot or cold towers. This change makes the fallback to the global average less disruptive.

## Key changes

- Updated cosmics calibration histogram inputs to use the HCAL ADC-to-ETower fields for inner and outer HCAL.
- Widened the calibration-factor reset range to only replace values below `0.01 × average` or above `10 × average`.
- Added warnings identifying bins whose factors are replaced.
- Added missing braces around affected control-flow blocks.

## Potential risk areas

- The changed histogram field names may affect compatibility with existing input files or calibration-tree formats.
- The wider acceptance range changes calibration and reconstruction behavior for extreme tower factors.
- No significant performance or thread-safety impact is expected.

## Possible future improvements

- Validate the new thresholds against representative hot/cold tower distributions and independent run3OO calibration samples.
- Add regression tests covering histogram-field selection and outlier handling.
- Consider making the thresholds configurable and documenting the expected input schema.

AI-generated summaries can contain errors; the implementation should be reviewed against the source and calibration validation results.

### DIFF
--- a/calibrations/calo/hcal_calib_year2/genFinalCalib/tsc_cos_merge.C
+++ b/calibrations/calo/hcal_calib_year2/genFinalCalib/tsc_cos_merge.C
@@ -54,10 +54,11 @@ void checkCos(TH2* h){
   
   for (int ie=0; ie<24; ie++)
     for (int ip=0; ip<64; ip++)
-      if (h->GetBinContent(ie+1,ip+1) < 0.01*avg || h->GetBinContent(ie+1,ip+1) > 10*avg)
-	std::cout << "Rescaling the calib. factor to a global average. Be cautious!" << endl; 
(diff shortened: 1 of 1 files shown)


pinkenburg:
This PR failed clang-tidy. This is caused by a bad bug in tsc_cos_merge.C. This shouldn't have been merged. Fixing this in the next PR

PR #1374, opened 2026-08-10
add OO noise scaling

pjwinnetou:
## Motivation / Context

Add pedestal noise scaling for `OO_COLLISION` beam configurations in calorimeter waveform simulations. The scale is `0.73`.

## Key Changes

- Add `OO_COLLISION` pedestal scaling in:
  - `common/G4_CEmc_Spacal.C`
  - `common/G4_HcalIn_ref.C`
  - `common/G4_HcalOut_ref.C`
- Use an exclusive `if`/`else if` chain for CEMC beam configuration handling.

## Potential Risk Areas

- The new scale can change simulated waveform pedestals and downstream reconstruction results for `OO_COLLISION` events.
- No I/O format, thread-safety, or performance changes are indicated.
- Validate the `0.73` value against detector conditions and compare reconstructed observables before and after the change.

- Centralize beam-dependent pedestal scale values to avoid duplication.
- Add regression tests for all supported beam configurations.
- Document the source and validity range of the `0.73` scale.

AI-generated summaries can contain mistakes. Review the implementation and validate the physics behavior before merging.

### DIFF
--- a/common/G4_CEmc_Spacal.C
+++ b/common/G4_CEmc_Spacal.C
@@ -349,15 +349,20 @@ void CEMC_Towers()
         std::string(getenv("CALIBRATIONROOT")) +
             std::string("/CEMC/LightCollection/Prototype3Module.xml"),
         "data_grid_light_guide_efficiency", "data_grid_fiber_trans");
-    // pedestal scale down for different beam configurations according to Blair
+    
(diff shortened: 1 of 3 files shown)


pinkenburg:
There will likely be an issue with embedding. The background even (OO or AuAu) does not set the species, the foreground sets it (for pythia8 it's pp with the beam xing). this can be changed since the embedding uses different macros but we need to agree on something

PR #1380, opened 2026-08-28
Fix sc use micromegas

hupereir:
This PR makes a few changes about how MICROMEGAS are used in (still default) CASeeder and Polyseeder workflow.
- SC_USE_MICROMEGAS is now set to false by default. (was true before), to make sure that TPOT hits are _not_ included in full trackfit (ACTS or GENFIT) by default. This is the intended behavior after discussion with Christof and other tracking experts. 
- Genfit now also uses SC_USE_MICROMEGAS to decide whether TPOT should be included in the fit or not. 
- in Tracking_Reco_TrackFit_run2pp() rather than hardcoding actsFit->setUseMicromegas() to false, restore using SC_USE_MICROMEGAS. This does not change the now default behavior, but would allow user to re-enable it by changing the flag in the steering macro; something that is not possible for now, and creates inconsistencies when SC_CALIB_MODE is turned on. @jdosbo can you confirm that you are ok with this change ? 
- Finally, in the steering Polyseeder macro provided by Mariia and SBU: re-add TPC-TPOT matching, and use SC_USE_MICROMEGAS to decide whether TPOT hits should be included in the full track fit. 

This should essentially make all macros consistent, while keeping the default behavior unchanged.

Comments welcome

## Motivation / Context

This update makes MICROMEGAS/TPOT handling consistent across the CASeeder and Polyseeder workflows. TPOT hits are excluded from full ACTS and GENFIT fits by default, while steering macros can enable them.

## Key Changes

- Set `G4TRACKING::SC_USE_MICROMEGAS` to `false` by default.
- Use `SC_USE_MICROMEGAS` to control MICROMEGAS inclusion in ACTS and GENFIT fits.
- Restore the steering-controlled ACTS setting in `Tracking_Reco_TrackFit_run2pp()`.
- Re-add TPC–TPOT matching in the Polyseeder workflow.
- Use `SC_USE_MICROMEGAS` for Polyseeder full-track fitting.

## Potential Risk Areas

- Reconstruction results can change when TPOT usage is enabled or disabled.
- No input or output format changes are expected.
- Performance may change when additional TPOT measurements are included in fits.
- Matching and fitting behavior should be validated for relevant data and simulation samples.

- Add regression tests for TPOT matching and fitting with both flag values.
- Document the expected physics and performance impact of enabling MICROMEGAS.
- Review default settings across all tracking workflows for consistency.

AI-generated summaries can contain mistakes. Contributors should verify these points against the code and validation results.

### DIFF
--- a/TrackingProduction/Fun4All_raw_hit_TPC_Matched_reco.C
+++ b/TrackingProduction/Fun4All_raw_hit_TPC_Matched_reco.C
@@ -342,11 +342,14 @@ void Fun4All_raw_hit_TPC_Matched_reco(
   silicon_match->set_crossing_deltaz_max(10);
   silicon_match->set_crossing_deltaz_min(0);
   silicon_match->set_test_windows_printout(false);
-  silicon_match->set_max_crossing_diff(10); // good for poly seeding case  
+  silicon_match->set_max_crossing_diff(10); // good for poly seeding case
   // these are for testing. and default to false. The seed matcher will choose the crossing that works best
   //  silicon_match->set_use_tpc_crossing_only(false);  // use crossing information from TPC SA seed
   //  silicon_match->set_use_silicon_crossing_only(false);  // use crossing information from silicon seed
   se->registerSubsystem(silicon_match);
+
+  // TPOT matching
+  Tracking_Reco_TpcTpotTrackMatching_run2pp();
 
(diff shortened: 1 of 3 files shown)


hupereir:
ping. Can someone push that, for cleanup ?

